PR #10, opened 2019-05-26
Feature/extract ps4 asset

Xeeynamo:
With this we will be able to extract the assets from PS4 versions of any remaster


Xeeynamo:
> hd-assets can also hold "/" in their filename. For example: voice/us/something.scd (can be found in a.fm files in kh2). You should check if the filename has any "forbidden" characters.

I just double checked with those files and the _ASSET_ folder will extract the given file in a sub-folder (eg. if the file 00test.bar contains an asset called foo/bar.bin, that file will be extracted to _ASSET_00test/foo/bar.bin). I think that this is the behaviour that we want?

PR #25, opened 2019-07-24
misc fixes on bar docs

GovanifY:
Excuse me in advance for this long PR and semi off-topic questions below, I tried to keep this as professional as possible.
 
* the type is only a uint16, it can contain a duplicate flag
* type 0x04 is not composed solely of VIF packets, albeit mostly, refer to my old tools for more details on their inner workings
* the zero value is a lookup address used at runtime

Keep in mind all of this was done through memory, so someone would have to double check this info i guess.

On another note don't you think something like a [kaitai](http://kaitai.io/) setup would be much more explicit for structures? That would allow you to get a parser for free on top of auto-generated docs ( I will avoid to start an argument about other technologies used by the project and only mention it can output C# )

Also any reason why most documented formats are not documented anywhere in this project? Not enough workforce? I am curious about this point since most formats have been extensively documented in the past.

By the way, unlike what you mentionned on twitter, AI, PAX, ANB and spawn points have been mostly reversed:
* Spawn points are available in kkdf2 tools from ages ago
* Several disassemblers have been made privately for KH2 AI. The AI is a small stack-based variable-width ISA with a huge number of "syscalls" if we can call them this way, which are by large the main reason AI reversing is such a pain
* ANBs has been mostly reversed by kkdf2 ten years ago again, with added information from soraiko and I
* While I did not worked myself on the PAX format, I remember some old effort about that in the KH groups.

To aid in reversing, RECOM BETA ISO is compiled in debug mode and some of its symbols can be reused. For example, the whole dpd logic basically has not changed _at all_ so this might have helped the reversing effort. Ghidra also has a work-in-progress [_decompiler_](https://github.com/beardypig/ghidra-emotionengine), which I would encourage you to look into since it can ease up immensely some reversing time(it is nowhere near perfect, but gets some basic tasks done). Ghidra SLEIGH specification is to keep in mind in cases like this, since you basically get a free decompiler when making a disassembler.

Anyways, sorry about the whole off-topic parts of the PR.
Cheers,
-G

### DIFF
--- a/docs/kh2/file/type/bar.md
+++ b/docs/kh2/file/type/bar.md
@@ -16,16 +16,17 @@ All the values are Little Endian in the PS2/PS4 Versions, while they are Big End
 
 | Offset | Variable Type | Description |
 |--------|---------------|-------------|
-| 0 | int32_t | The identifier of the file (Should be always 0x01524142) |
-| 4 | int32_t | The sub-file count of the BAR File. |
-| 8 | int32_t | Always zero. Has no known effects.
-| 8 | int32_t | Unknown. Often 0. Some mset files set it to 1 or 2.
+| 0 | char[4] | The identifier of the file (Should be always 0x01524142) |
+| 4 | uint32_t | The sub-file count of the BAR File. |
+| 8 | uint32_t | Always zero. Padding for a lookup address at runtime. 
+| 12 | int32_t | Unknown. Often 0. Some mset files set it to 1 or 2.
 
 ### BAR Entry
 
 | Offset | Variable Type | Description |
 |--------|---------------|-------------|
-| 0 | int32_t | The sub-file's type (List given below) |
+| 0 | uint16_t | The sub-file's type (List given below) |
+| 2 | uint16_t | Duplicate flag (if the entry is already included in the BAR) |
 | 4 | char[4] | The name of the sub-file. Empty characters are registered as 0x00.|
 | 8 | uint32_t | Sub-file's offset/location. Must be divisible by 0x10. Can be padded. |
 | 12 | uint32_t | Sub-file's size. Everything that comes after [ offset + size ] will be ignored for that specific sub-file.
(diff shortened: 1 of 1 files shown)


Xeeynamo:
Hey Gauvain!

The documentation is not exactly in line with the core libraries (eg. the duplicate flag in BAR files). This is something that we definitively have to fix.

I personally worked for just few hours on both AI and PAX from scratch, as I do not have any base to work with. But I would be glad to not re-invent the wheel and speed-up the development! For ANB (and mostly 3D stuff), it is something that I personally do not want to touch and I want to wait to have other people to get their hands on.

I really would like to use Ghidra, but my attempts to use it with PS2 ELFs were not great. Kaitai seems to be a more production-ready version of BinaryMapper, which is interesting.

Thanks for the suggestions, the PR itself and all the details that comes with it. I am merging to `master` branch straight away!

GovanifY:
Hey,

Ghidra wise, while beardypig plugin isn't as polished as IDA if nobody ever does the yak shaving we'll be stuck in such a situation. I understand that currently IDA seems like the easiest choice and it most likely will stay this way since very few people still work on the PS2, nevertheless it is still something to keep in mind if you ever want to learn more about PS2 co-processors operations and improve the plugin.

Kaitai wise, it's waaaay more mature than BinaryMapper. Shameless plug but years ago I created some kaitai definitions which you can see [somewhere in this post](https://govanify.com/post/re-3d-model-mdlx/) for KH2. If we really want auto-generated documentation then we could make use of the doc parameters of kaitai flags and have an explaination in proper english. It is really well designed and you can do pretty much whatever you think about with it.

I'm willing to help later for kaitai parsers and documentation but totally unwilling to help with any non-portable WPF-based GUI. You know my thoughts on the technology you're using and I believe that when I am unable to even _use_ the tools that I'd work on, there would be an obvious issue for both you and I.

Oh and simply curious but why aren't you interested about the 3D formats? They're one of the most convoluted and fun formats of KH2 to reverse, you're missing something haha.

While I do not have a discord account anymore(I will contact you in some other way when I am ready to help) your discord invite link in your contribution guide has expired, you might want to fix this :)

PR #30, opened 2019-08-17
Create pam.md

Disgustor:
Initial Birth by Sleep PAM doc


Xeeynamo:
Looks good! I suggest though to add the PAM link in `docs/bbs/index.md` to make it easier to reach

PR #36, opened 2019-09-02
Added many lines to ./docs/kh3/pawns.md

Vladabdf:
Includes PCs, NPCs, Enemies, and Weapons. Probably not complete as many spaces are blank, many duplicates marked with (?), some marked by #, such as CDA Agents #1, #2, and #3.


Xeeynamo:
Are you sure that the path of weapon is that? The rest looks good, amazing job!!!

Vladabdf:
I'm not so certain, I just copied the line for the most part and changed what I assumed would be different as I currently have no way of testing it

PR #43, opened 2019-11-09
KKDF2's Docs simplified

Tomk1004:
All credit goes to Mr. Kenjiuno for his docs, I have simplified the first parts of it as those parts are concrete fact, while the things closer to the end of his docs get into speculation, though they may be accurate. The files added are the same, I've added the pdf in case you don't want to download the zip


Xeeynamo:
Thanks for the pull request and for the documentation!

I was thinking if it would be the best to have the documentation in an editable format (html or md) rather than a PDF? So anyone can expand it later on.

Let me know if there are some limitations that prevents you to do so!

Tomk1004:
https://drive.google.com/file/d/1RFVVDo8lqKXGeDk8oC1zOh4Kr3i6Y_WY/view?usp=sharing link to the html file

Xeeynamo:
Looks great! I did some small changes (converted from html to markdown) and created another pull request. Please have a look before the merge https://github.com/Xeeynamo/OpenKh/pull/45

PR #55, opened 2019-12-25
more ai doc

GovanifY:
I'll be quite honest, I'm not sure what you want exactly with my AI doc so i just made this very barebone PR adding my licenses files.
All my documentation code itself is under the MIT license while the output is under the CC-BY-4.0 license.

You can include any part of the project with those licenses that you care about and modify it and all assuming you respect the license terms.

I'll just ask you, in a not legalese manner, two things:
The first is to contact me if you want to change the documentation. The pdf has a link to a form to report new opcodes/syscalls/erratas. I'd like to avoid segmentation of this manual and as such, if possible, make it so that i release numbered versions myself and you keep a copy of the numbered releases. You are of course free to modify the documents yourself, according to my own license terms, so none of this is actually legally binding.

The second point is to avoid as much as possible WPF-based GUIs which hinder the portability of your tools. Again, nothing formal, but while I understand porting older tools would take some work I do not think writing any new GUIs on another UI toolkit would be that much of an hassle to deal with.

Either way you're free to fork and edit my repos with those license files, tell me if you have any more concerns


GovanifY:
Re licenses at the root of the directory: This does not necessarily mean that.
Licenses are applied on a suggestion basis. If you mention in the readme that X LICENSE applies for X and Y applies for Y then you can put it at the root of the directory without any issues. As my work only applies to AI for now I am fine with moving them to the ai subfolder if you really want to though.

For the documentation format itself: you're on your own. My entire documentation is made in LaTeX and is unlikely to change as PDFs are a standard for ISAs. You can try to make a markdown formatter for it if you're feeling like it. The CC license only applies to any generated document that includes my documentation, so your markdown would fall under the CC-BY license and your markdown generator under MIT if that's what you're wondering.

Xeeynamo:
I have an idea then:

What if we just use a script (GitHub Action, Azure Pipeline, etc.) to generate the markdown from your repository to the documentation?

The reason is because I would like that, if we have to add new documentation to the AI, we would directly open a pull request to your repository to centralise the changes. Plus, OpenKh docs will be automatically be updated bases on the latest commit of your repository.

Does it sounds reasonable?

GovanifY:
It does sound perfect to me! Unfortunately I'm heading over to germany so I won't be able for a few days to look at your webhooks.

By the way, for the future of kh2ai I might need some helping hands if we want something done quickly. Do you have a preferred way to talk with me? 
I honestly believe it should be a matter of a few weeks of full-time work at best(so assume some months of random work) and I can give you resources that should ease up reversing.

Xeeynamo:
@GovanifY I am trying to build the documentation, but I get an error about `blindtext.sty` missing when I try to run `htlatex` or `pdflatex`. You can see the error on the CI/CI pipeline: https://ci.appveyor.com/project/xeeynamo/openkh-kh2ai/builds/29967629

Surprisingly it works fine using the package `texlive` on Arch. But with a Debian/Ubuntu image I got that error even if I am installign the packages `texlive tex4ht`. Do you have any idea of what the cause would be? Maybe a missing package or some kind of mis-configuration?

GovanifY:
@Xeeynamo `apt install texlive-latex-extra` should do the trick

Xeeynamo:
It does, thanks!

That said, I created the pipeline to automatically generate the KH2AI docs in markdown format: https://ci.appveyor.com/project/xeeynamo/openkh-kh2ai/build/artifacts

The script I am using to do it is: https://github.com/openkh/openkh-kh2ai/blob/master/Makefile

The optimal solution would be to create a pull request from the artifacts, but I had to build a webhook and it's not worth the time :P

Since we found this other solution, are you okay to close this pull request without a merge?

Also be free to contact me at xeeynamo#5057

GovanifY:
Yup, saw your script in the CI above! While the technology involved is pretty counterintuitive to say the least(I believe it would have been much saner to simply parse every `\ISA` line from my LaTeX file) the end result is actually really clean!

Definitely fine with closing this PR, as long as my licenses are respected with derivatives of my work.

Also I'll definitely contact you later! Have a nice day

Xeeynamo:
> Definitely fine with closing this PR, as long as my licenses are respected with derivatives of my work.

I need your confirmation regarding that :)

GovanifY:
Include my MIT license to any generator code regarding my documentation and the CC-BY to the documentation output.
As a thumb note: if code > MIT otherwise CC-BY

PR #57, opened 2019-12-30
First Tool Documentation - KH2 Text Editor

Vladabdf:
Extensive detail on how to use the text editor. Many functions aren't currently listed either because their types are marked as unknown or I simply didn't know their purpose or how to use them. It covers all the main bases though. This document should be totally usable for the majority of people that wish to change text in the game.
Includes proper markdown formatting and example images littered throughout.


Vladabdf:
I think that's an excellent idea for this program and future ones.

PR #64, opened 2020-01-05
Fix certain errors with Item.cs

Topaz-Reality:
Because I could not resist.

### DIFF
--- a/OpenKh.Kh2/System/Item.cs
+++ b/OpenKh.Kh2/System/Item.cs
@@ -6,35 +6,37 @@ namespace OpenKh.Kh2.System
 {
     public class Item
     {
-        public enum Type : byte
+        public enum Type : ushort // Declaring this as a byte is a mistake you do NOT want to make. Trust me, I have been there.
         {
-            Consumable,
-            Boost,
-            Keyblade,
-            Staff,
-            Shield,
-            PingWeapon,
-            AuronWeapon,
-            BeastWeapon,
-            JackWeapon,
-            DummyWeapon,
-            RikuWeapon,
-            SimbaWeapon,
-            JackSparrowWeapon,
-            TronWeapon,
-            Armor,
-            Accessory,
-            Synthesis,
-            Recipe,
-            Magic,
-            Ability,
-            Summon,
-            Form,
-            Map,
-            Report,
+            Consumable = 0x00,
+            CampItem = 0x01,
+            MegaConsumbale = 0x0200,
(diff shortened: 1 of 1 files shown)


Vladabdf:
`            [Data] public byte FireVulnerability { get; set; } // These show "Vulnerability"
            [Data] public byte IceVulnerability  { get; set; } // Not "Resistance"
            [Data] public byte LightningVulnerability  { get; set; } // They are in percentages
            [Data] public byte DarkVulnerability  { get; set; } // "0x64" to "0x00"`
While correct, values span from 0x00 all the way to 0xFF, with 0x64 being "no change" from regular damage. For example, Petite Ribbon changes those vulnerability values to 0x5A. Through testing changing them to 0xFF, Sora becomes far more vulnerable to those damage types.
There are also types for "physical", "light", and "neutral magic" resistances, though I do not know exactly which bytes in the file are attributed to them, as there is some uncertainty amongst modders as to which attacks are all affected by the "light" and "neutral magic" values. Reflect and Explosion are certainly "neutral magic" while I suspect Roxas' orbs are "light". More testing needed to verify them explicitly.

Xeeynamo:
The `Item.Type` is 100% a byte, according to how the game reads the data. If you look to the routine `1a35d0` (KH2FM version), the assembly instruction `lbu t6,0x2(v0)` (where `v0` is the item entry) clearly reads the type as a byte.

Good point about element vulnerability rather than resistance. It is still a bit obscure to me how the game reads it as it still needs to be implemented in the editor.
Would be possible for you to move those comments in the code to the documentation of the file format? I think it would be more clear.

Finally, your pull request is breaking the build, due to the editor `OpenKh.Tools.Kh2SystemEditor` that is unable to pick your changes.

Rikux3:
Isn't `0x2` in the assembly instruction the buffer size? If that's the case, it's reading a short and not a byte. Or am I reading it wrong?

Xeeynamo:
@Rikux3 Nope. When in assembly you find `0x20(a1)`, it means to go to the pointer `a1` and move 0x20 bytes from there.
In MIPS assembly, you can recognise if it is 1, 2, 4 or 8 bytes if the instruction is `lb`, `lh`, `lw` or `ld` (where `l` is load, `s` is store, `u` is unsigned; eg. `swu` stores 4 unsigned bytes`)

Rikux3:
I assume we can close this one as well?

Xeeynamo:
Close due inactivity. The build was broken anyway.

PR #67, opened 2020-01-13
Add japanese support to KH2 Text Editor

Xeeynamo:


Rikux3:
Support for special characters like `XI` or `III` isn't implemented right?

Xeeynamo:
That is a really strong point. There is no support for those characters. I will fix it in few hours.

Xeeynamo:
Are there not III, XI and XII in the JP font, am I right?

Rikux3:
Doesn't look like it . I've attached the whole font map.

Vladabdf:
I spotted III in the provided image. Fourth row, center-right.

Rikux3:
@Vladabdf The characters in the first "block" (top-left) are interpreted as strings, so these are already implemented.

Xeeynamo:
Done!

Rikux3:
Opening text 1447 in sys.bar results in a crash in JapaneseSystemEncoder.cs::AppendEncodedChar
grafik

Rikux3:
`XIII` also appears to render wrong

grafik

Xeeynamo:
I swear that now is working 👀

Rikux3:
I'm okay with that. The progress made in this branch is already big enough.

PR #83, opened 2020-02-24
BBS CTD and Font support

Xeeynamo:
This pull request adds the code to parse and save a CTD, plus read support for BBS fonts and a bare bone text editor. It also allow the Image Viewer app to load and display the content of a font file.

The CTD parser is complete and the docs too. There are few unknown fields and unexplained logics that the game uses (like chaining multiple dialogs for a NPC). Also the CTD editor uses the CTD layout correctly, but it does not allow to modify it. Same for the font editing functionality built-in, where there is no Save support for write back the changes to a font file.

Editing the dialogs is fully supported and the Layout is almost fully documented too, so it will be quite straightforward to add that functionality in the editor in a second iteration.

Japanese text can not be rendered too as a similar functionality on KH2 Text Editor needs to be implemented as well.

I will create issues for anyone of those missing features, so anyone can pick them up at any time.


Xeeynamo:
I created the following issues for the optional unfinished matters that this pull request does not cover: https://github.com/Xeeynamo/OpenKh/issues/84 https://github.com/Xeeynamo/OpenKh/issues/85 https://github.com/Xeeynamo/OpenKh/issues/86 https://github.com/Xeeynamo/OpenKh/issues/87

PR #95, opened 2020-05-07
Objentry additions

Rikux3:
Updated some previous unknown fields and the documentation


Rikux3:
`[Description]` and `[Description("")]` show the same in the UI = an empty string. It seems that the Attribute isn't using the enum name as a fallback. I have removed the empty descriptions nevertheless because they looked ugly.

I didn't like the way i've implemented the `Clone` function before, because it didn't recognize changes in the structure of the object. I've also thought about implementing the `IClonable` interface but in the end i've liked the Reflection approach better (although it can have side effects).

Xeeynamo:
Looks good 👍

PR #101, opened 2020-05-15
OpenKH Engine (placeholder)

Xeeynamo:
So far, this is the biggest pull request of the project.

I added `OpenKh.Game` project, a MonoGame executable that uses the various library to "simulate" Kingdom Hearts II game engine. Let's go through the features:

* IDX/IMG reading: can load game assets from `KH2.IDX` and `KH2.IMG`. If you want to make it working from the extracted content, unpack the game there.
* Region detection: checks the existence of `menu/{region}/title.2ld` to understand from which region the game belongs to. Useful to load any language asset from any PS2 game build.
* Title Screen support: can navigate through it. When you select an option, it will take you into a map. The graphics might be buggy on non-FM versions of the game.
* Map support: Use W/S and arrow keys to navigate through the map.
* Debug support: Press TAB to access to the in-game debug functionalities. Press Left and Right to scroll through the menu. L is CIRCLE, K is CROSS.
* Font and message: The game loads the font table and `sys.bar`, printing text as close as possible as the game does. Note that the encoding used is from the international version, meaning that the text will be random if you load JP or un-patched FM builds.
* Objentry support: detect all the game models and loads some of them without problems. Thanks to @Rikux3 .
* `place.bin`: show all the map names.
* 00battle/03system: it's loading part of those files. Thanks to @Rikux3 .

Now, **before accepting this pull request**, there are some refactors I want to perform to consolidate the code between the tools and the game engine. Doing so, they will share the same code.

Also, since I never understood how VIF works, I shamelessly copied @kenjiuno code to unpack them. I would like to properly refactor and document that portion of code, but I have no idea where to start. If he approves to use its code here, then I would say to ignore this issue for now. Keji code can be found in `OpenKh.Engine/Parsers/Kddf2`.

image
image
image


Vladabdf:
I'm loving how well this is turning out and I'm testing every few commits to make sure nothing breaks. I haven't reviewed + accepted the merge request yet though because as you said, biggest pull request yet, so we gotta make sure it's juicy enough to include in master. :sunglasses:

PR #106, opened 2020-05-22
Upgrade from .NET Framework 4.7.2 to .NET Core 3.1

Xeeynamo:


Rikux3:
How does this influence the runtime requirements? When you publish a release, are all dependencies included or do you have to have the .NET Core 3.1 runtime installed?

Xeeynamo:
I am afraid you need to have .NET Core 3.1 runtime installed.

Vladabdf:
I'm happy with this change; even on many builds of Windows 10, .NET 4.7.2 isn't installed by default and must be enabled through Windows settings. Because of that, it's not more difficult to install .NET Core. To compensate, I changed the readme to include a link to download the necessary runtime for all major OSes.

Xeeynamo:
.NET Framework 4.7.2 is preinstalled in all Windows 10 builds past 2018: https://docs.microsoft.com/en-us/dotnet/framework/get-started/system-requirements

@Rikux3 made a strong point though, since it is possible to include the .NET Core runtime when shipping a new build to GitHub Release. Although it will increase the size of the download from 3.5MB to over 150MB.

But another good point is that all the command line applications requires .NET Core 3.1 anyway, so it might be worth to consolidate the framework runtime across all the projects.

Rikux3:
I don't really like how .NET Core applications ship. On one hand, it's cool that all dependencies ship and you dont have to install anything. On the other hand, it's blows up the file / archive size massively. Best example I had in the past was a small tool about 200 LoC which was 50 megabytes with .NET Core and under 1 megabyte with .NET Framework 4.7.2.
Moving to the .NET Core architecture is definitely the right step, but this really bugs me.

Xeeynamo:
File size bugs me as well. But it is less a pain 

We have two options here if we plan to go forward with this pull request:

1. We ship the framework with OpenKH, so users will never be bothered at installing new versions of runtimes. But again, expect a 100/150MB of file size increase and a massive amount of DLLs in the same folder where all the tools and the game are.
2. We specify that there is a runtime requirement and users have to download it. Size and content will be considerably smaller, but we also have to be careful if we want to upgrade to .NET 5 or .NET 6 because they have to install the new runtime every time.

Vladabdf:
I've seen projects switch runtimes every single time a minor version releases and I've never really seen major community backlash to accompany it. That being said, theoretically as long as we don't make these jumps frequently, it shouldn't be a big deal. I do side with specifying and having people download and install the runtimes themselves though. However, we do need more input to decide whether the pros outweigh the cons.

Cibbi:
If it may be of any help, if you try to launch a net core application without having the runtime the application will actually say that and if i remember right it should also redirect to the download of the runtimes (not 100% sure, would neet to test it). 

So i think the second option is the better one.

Xeeynamo:
@Cibbi You're right: that's really good!!!

image

Vladabdf:
So I guess this is the solution we needed then. :)

Xeeynamo:
image

This is the message thrown when I installed a later runtime than 3.1. It redirected to the following page, which is very well structured and brain-less page: https://dotnet.microsoft.com/download/dotnet-core/3.1/runtime/?utm_source=getdotnetcore&utm_medium=referral

PR #115, opened 2020-05-25
Updating the program to run tests.

Soraiko:


Xeeynamo:
Can we have more info about the the intent of the changes you proposed?

Soraiko:
> Can we have more info about the the intent of the changes you proposed?

Sure. I changed one file, Program.cs, to make it point the directory "OpenKh\OpenKh.Tests".
That's pretty much all.

PR #117, opened 2020-05-26
KH2: Japanese EVT text decoder and encoder

Rikux3:
Double-checked with kenjiuno's branch to be sure.


Xeeynamo:
The test `OpenKh.Tests.kh2.MsgEncoderTests.DecodeJapaneseEventTextCorrectly` is broken, probably due to a false positive since the Event encoder is different than the System one.
Can we double-check before merging into master?

kenjiuno:
@Rikux3 Thanks for pushing this!

I have changed ゎ to わ:

before:

lwa

after:

wa

System font (and XIII) seems to be ok!

sys-20402

Event font seems to be ok!

po-4663

kenjiuno:
I suggest comment out the test failure part simply (and merge pull request).
https://github.com/Xeeynamo/OpenKh/pull/117/files#diff-ce58775d852863c5b1fa7408e15b6ba0R251

It means `JapaneseEventTable._table8` won't be used/verified, for now.

I think it may take much longer, if I review TextDecoder system's WrapTable algorithm in this pr...

Xeeynamo:
According to #120, MsgTool is broken for Japanese text. I am using the content of this pull request to add support to it. You can now specify `-e jpsys`, `-e jpevt` or `-e eusys` based on the region.

kenjiuno:
Thanks @Xeeynamo I have checked MsgTool too!

```bat
OpenKh.Command.MsgTool.exe -i H:\KH2fm.yaz0r\msg\jp\sys.bar  -e jpsys
```

```
...
  <message id="20400">
    <text>集積の庭園に集められた</text>
    <complex value="XIII" />
    <text>機関の再現デ–タをすべて倒した証</text>
    <text>ソラに何か变化が…?</text>
  </message>
...
```

jpsys will be ok 👍 

```bat
OpenKh.Command.MsgTool.exe -i H:\KH2fm.yaz0r\msg\jp\po.bar  -e jpevt
```

```
...
  <message id="4663">
    <text>プ– 久しぶり!</text>
    <delay value="50 00" />
    <clear value="" />
    <text>ハチミツたくさん食べてるかい?</text>
  </message>
...
```

jpevt will be ok too 👍 

Wait!

At `プ–` (Pooh) and `デ–タ` (data), `–` is hyphen. I think it will be [Chōonpu](https://en.wikipedia.org/wiki/Ch%C5%8Donpu)
I'll push fix for this...

PR #118, opened 2020-05-26
Add custom shader to simulate PS2 behaviours

Xeeynamo:
This fixes some of the odd texture rendering glitches caused by the inability of emulating the "Texture Wrap Mode" as PlayStation 2 internally does.

BEFORE
before

AFTER
after

### DIFF
--- a/OpenKh.Game/Content/KingdomShader.fx
+++ b/OpenKh.Game/Content/KingdomShader.fx
@@ -9,6 +9,7 @@
 
 matrix WorldView;
 matrix ProjectionView;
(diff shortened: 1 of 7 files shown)


Rikux3:
Out of curiosity: What causes these "smudges" in the skybox? I've found them on another map too. 
grafik

Xeeynamo:
I think it is a regression caused by the custom VIF interpreter I wrote before merging the engine into master. I am going to confirm later in the day if it is caused by the a regression of this pull request or somewhere else.

PR #107, opened 2020-05-22
BAR editor improvements

Xeeynamo:
Add CLI tool for BAR files, drag&drop support, file extension for any type, fixes when writing back or creating a BAR and reintroduced back the possibility to open other tools from the BAR editor.

PR #119, opened 2020-05-27
Fixes countless bugs that risked to damage files changed from #107

Xeeynamo:
The Pull Request #107 renamed `EntryType.Binary` to `EntryType.List` and added a `EntryType.Binary` for a different kind of value. 
But most of the tools did not picked up the change and are still using the `EntryType.Binary`  rather than `EntryType.List`. This change breaks tools and game engine and the pull request aims to fix it.


kenjiuno:
Thanks for fixing!
OpenKh.Tools.Kh2TextEditor
OpenKh.Tools.Kh2SystemEditor
OpenKh.Tools.LevelUpEditor
... will be ok on load/save.

PR #121, opened 2020-05-27
[wip] mset early&initial support on OpenKh.Game

kenjiuno:
<s>Segments → `Func<double, Segment[]> GetSegmentsByTime`</s>

I'm not good at extending high level library like Engine.

anim_demo


kenjiuno:
`.mset` is loaded inside FromEntity. It would be better on loading separately or upper layer.
https://github.com/Xeeynamo/OpenKh/blob/4f1add3f22307a9437dfb3493bf65e51ae6a393d/OpenKh.Engine/Parsers/MdlxParser.cs#L88

kenjiuno:
Both `.mdlx` and `.mset` binaries are required because KH2's PS2 code needs (or eats) it.

Xeeynamo:
This is a very good start as a proof of concept! Are you planning to refactor it and make it ready to be merged into master?

kenjiuno:
> This is a very good start as a proof of concept! Are you planning to refactor it and make it ready to be merged into master?

Thanks for checking. I got animation implementation idea by googling MonoGame.

[Model](https://docs.monogame.net/api/Microsoft.Xna.Framework.Graphics.Model.html#properties) class has native support of `Bones` and `Draw` using matrices. So current MSET support code will be more simple a little.

kenjiuno:
Model and [ModelBone](https://docs.monogame.net/api/Microsoft.Xna.Framework.Graphics.ModelBone.html)

In my interpretation, ModelBone is to be used to connect body and parts (having parent and child relationship).

image

[SkinnedEffect](https://docs.monogame.net/api/Microsoft.Xna.Framework.Graphics.SkinnedEffect.html) is useful if we pass SubModels into GPU like KH2 does in VPU.

image

Untransformed vertices (as vertex buffer) and matrices (as constant buffer) can be processed on SkinnedEffect's specific vertex shader.
SkinnedEffect supports up to 72 matrices at once.
WeightsPerVertex is 1, 2, or 4.
An issue is that **KH2 uses WeightsPerVertex = 5** at most!

Xeeynamo:
I am happy to see that the Model Matrix multiplication is now performed in the GPU! That is a good practice for 3D applications since it increases the performance. 

I did not looked yet to your implementation, but usually the original meshes are stored into a Vertex Buffer. Then for each mesh, you send to the shader the Matrix animation and draw only the portion of the vertex buffer associated to the mesh it needs to be transformed.
This is done to obtain great performance, but it is not needed to OpenKH engine yet. Sorry if my explanation was not clear.

I am really excited for this pull request 😁

Xeeynamo:
Hello @kenjiuno 😃are you still working on this? If you need to pair with someone to carry on with this, we can both work together! If not, take your time. We look forward on this pull request!

Vladabdf:
Closed in favor of #256

PR #122, opened 2020-05-27
Fix model load

Xeeynamo:
This fixes a couple of bugs that prevent some 3D models to be loaded correctly. This patch will allow to load all of them, regardless their "type".


Vladabdf:
Certain models appear to crash the game. From what I can tell, it's mostly hi-poly models. Examples being ACTOR-SORA-H and H_PO010. However, P_EX110_BTLF, which is a low-poly model, also crashes the engine. What models do load though, load correctly!
image

kenjiuno:
> Certain models appear to crash the game. From what I can tell, it's mostly hi-poly models. Examples being ACTOR-SORA-H and H_PO010. However, P_EX110_BTLF, which is a low-poly model, also crashes the engine. What models do load though, load correctly!

Oh, it is refactored code that I have pushed around last week. I'll check it later!

kenjiuno:
Ok, ACTOR-SORA-H will be loadable now.

Xeeynamo:
@Vladabdf I confirm that @kenjiuno fixed the issue of loading high poly models!
image

Issue #124, opened 2020-05-30
Autodetect unstripped ReMIX files


Vladabdf:
I think this would be a useful thing to have, specifically for the game engine so far. It would also be a good start to finally interweave our tools together. Upon launching the game engine if it detects ReMIX files but notices they are not stripped, prompt the user to strip them before finally starting the game.

Xeeynamo:
easy done #126

Rikux3:
Fixed with #126

PR #126, opened 2020-05-31
HD Asset detection

Xeeynamo:
This fixes the issue #124, where it was not possible to load PS4 remastered assets without stripping out the HD content.

#124 suggests to prompt the user to strip the HD stuff, which implies that the game files would be modified. This pull request takes a step further and just let the game engine to work without modifying them. The performance impact of this pull request is that now the game uses more ram for PS4 files, since all the HD textures will be loaded in memory as well.

When the HdAsset support for PS3 will be implemented, there will be potentially no changes to the game engine to make it working for PS3 too.

Finally, there are minor API changes:
* `HdAsset.New()` has been deleted, since it would make more sense to just invoke `new HdAsset()`
* Adds `HdAsset.IsValid`, which tries to detect if a PS4 ReMIX HD asset file is valid.

### DIFF
--- a/OpenKh.Common/Archives/HdAsset.cs
+++ b/OpenKh.Common/Archives/HdAsset.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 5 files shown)


Rikux3:
Works perfect with 1.5+2.5 JAP.

PR #127, opened 2020-05-31
KH2 ISO injection and IDX+IMG compression

Xeeynamo:
Adds support to inject files into an ISO file, which is very useful when you want to quickly test a change you made to the game files. 

At the moment you have to specify the offset of both `KH2.IMG` and `KH2.IDX` of the ISO you want to modify, but I default the values to the ISO I am using: a patched Kingdom Hearts II Final Mix using Crazycat00's English translation. It should work with the japanese Final Mix version too. If not, you have to find the offsets.
This will be automated once the format ISO9660 will be implemented or used. If we can find an open source library it would be even better.

I am personally using this new functionality using the following script to automate some stuff I am recently working on:
```
#!/bin/sh
rebar ./export/map/tt00/tt00.map.json -o ./export/map/tt00_mod.map
injectiso ./KH2FM-EN.iso ./export/map/tt00_mod.map map/jp/tt00.map
../ps2-emu/pcsx2 ./KH2FM-EN.iso
```

This is also using a compression algorithm I developed [9 years ago](https://github.com/Xeeynamo/KingdomHeartsPatch/blob/master/libKh/IDX.cpp#L52) and it is the only one I am aware of that have a decent compression ratio and it does not break the de-compression routine of the game in very specific edge scenario. This will allow to be a step closer to have a tool in OpenKH that is capable to repack the IDX+IMG. But since it already exists, it is not a priority and it will not be part of OpenKH any time soon.


Xeeynamo:
@Rikux3 I will default it to the un-patched ISO then. Thanks for checking!

Vladabdf:
Just for future reference, the reason the offsets don't align is because of the modified 03system.bin in the English patch, which is one of the earliest files to load. Applying the English patch without that modified 03system will work perfectly fine with this methodology.

PR #128, opened 2020-05-31
00battle Editor and Docs

Rikux3:
There are still sub-files missing, but we did the same with the System editor, so i think we should still merge this.


Rikux3:
Sorry about that. As i've mentioned earlier, i might have made errors while rebasing this branch. Should be fixed now.

Xeeynamo:
All good! The tool is very useful and it gives loads of modding features we were previously unaware of.

There are some tweaks that would be ideal to see in the future. But let's take them as just a wish-list:
* Abilities could have their real name rather than just an ID
* Group characters on the Bonus level tab
* Show the EXP required to level up as a linear graph and avoid to insert a lower value than the previous level.

PR #141, opened 2020-06-06
Try to load plugin from `.dll` instead of `.exe`

kenjiuno:
- Try to load plugin from `.dll` instead of `.exe`
- Xeeynamo/XeEngine.Tools.Public#2 is also needed to succeed initialization of `ImageViewerViewModel._imageFormatService` which throws `TypeInitializationException` (on resolving `OpenKh.Bbs`).

### DIFF
--- a/OpenKh.Tools.BarEditor/Services/ToolsLoaderService.cs
+++ b/OpenKh.Tools.BarEditor/Services/ToolsLoaderService.cs
@@ -32,7 +32,7 @@ namespace OpenKh.Tools.BarEditor.Services
 			}
 
 			var toolModule = Plugins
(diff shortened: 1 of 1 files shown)


Xeeynamo:
Unfortunately it still does not work :( . What I am trying is to build the solution using `build.ps1` and then trying to open an image from the BAR editor. If you have a moment, could you investigate more? Maybe I am doing something wrong. Thanks!

kenjiuno:
> Unfortunately it still does not work :( . What I am trying is to build the solution using `build.ps1` and then trying to open an image from the BAR editor.

Ok, thanks for checking and merging XeEngine.Tools.Public!
I have updated sub module. Please try git pull from OpenKh, and run `build.ps1`. It will work now.

PR #142, opened 2020-06-06
Fix rendering for meshes that uses transparent textures

Xeeynamo:
The image below speaks itself

goodbye cyan background


Rikux3:
Tested with several maps, looks good!

PR #144, opened 2020-06-07
Implement key repeats on debugMenu.

kenjiuno:
This pull request adds key pressing pulse in debugMenu.

repeats

It uses actual SPI_GETKEYBOARDDELAY/SPI_GETKEYBOARDSPEED values obtained from [SystemParametersInfoA function (winuser.h) - Win32 apps | Microsoft Docs](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfoa) Win32 API.
You can change it from `control.exe keyboard`

### DIFF
--- a/OpenKh.Game/Infrastructure/InputManager.cs
+++ b/OpenKh.Game/Infrastructure/InputManager.cs
@@ -13,8 +13,8 @@ namespace OpenKh.Game.Infrastructure
 
         public bool IsDebug => repeatableKeyboard.IsKeyPressed(Keys.Tab);
         public bool IsShift => repeatableKeyboard.IsKeyPressed(Keys.RightShift);
-        public bool IsDebugRight => repeatableKeyboard.IsKeyPressed(Keys.Right);
-        public bool IsDebugLeft => repeatableKeyboard.IsKeyPressed(Keys.Left);
+        public bool IsDebugRight => keyboard.IsKeyDown(Keys.Right) && !prevKeyboard.IsKeyDown(Keys.Right);
+        public bool IsDebugLeft => keyboard.IsKeyDown(Keys.Left) && !prevKeyboard.IsKeyDown(Keys.Left);
         public bool IsDebugUp => repeatableKeyboard.IsKeyPressed(Keys.Up);
         public bool IsDebugDown => repeatableKeyboard.IsKeyPressed(Keys.Down);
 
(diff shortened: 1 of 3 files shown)


kenjiuno:
Ok, Linux... I'll check it for a while.
If I cannot find better way, I'll remove Win32 P/Invoke part and use fixed values.

Xeeynamo:
I made some research. Apparently the best way to do it is to store each pressed key with a timestamp of when the key has been pressed: `Key[] Keypool = new Key[MaxKeyCount];`
Where each index of that array can be a key:
```
const int KeyTriangle = 2;
const int KeySquare = 3;
...
Key GetKey(int key) => KeyPool[key];
```

A key can be represented by this:
```
struct Key {
    public bool IsPressed;
    public float Timer;
    public int PressCount;
}

where `Timer` can be updated using the deltaTime from `Update`.

There could be functions like the following:
```
bool IsKeyPressed(int key) => GetKey(key).IsPressed;
void PressKey(int key) => KeyPool[key] = new Key { IsPressed = true };
void ReleaseKey(int key) => KeyPool[key] = new Key { IsPressed = false };
void UpdateKey(int key, float deltaTime) => KeyPool[key].Timer += deltaTime;
```

And finally the function that would manage the repeats:
```
const float MinimumRepeatTime = 1.0f; // seconds
const float ContinuousRepeatTime = 0.25f; // every 250ms

bool IsKeyRepeat(int key)
{
    if (!IsKeyPressed(key))
        return false;
    timer = KeyPool[key].Timer;
    if (timer < MinimumRepeatTime)
        return false;
    timer -= MinimumRepeatTime;
    if (timer < key.PressCount * ContinuousRepeatTime)
        return false;
    key.PressCount++;
    return true;
}
```

I do not know if the code I written above is correct or it compiles. I written it in GitHub. But it should more or less replace what `SystemParametersInfo` is doing internally?

I do not want to take over your work, so I will leave you complete freedom. But please let me know if you have better ideas 😄. I trust your judgement and skills.

kenjiuno:
Sorry for late reply. I'll check your code later!

kenjiuno:
Ok thanks for advice @Xeeynamo !
It is better now!

And I have removed Windows specific P/Invoke `SystemParametersInfo`.
Timing setting (MinimumRepeatTime/ContinuousRepeatTime) will be better to be Engine specific parameter rather than OS dependent.

Xeeynamo:
It looks really good! Congratulations 🎉be free to merge it whenever you want

kenjiuno:
Thx!

PR #150, opened 2020-06-08
Change location for game content

Xeeynamo:
Previously you had to put all the game files into the same directory where binaries were located. I changed it to fetch game data from the folder "data".
It is also possible to specify the game data directory through command line, so you can just invoke `dotnet OpenKh.Game.dll /mnt/disc0`.


Xeeynamo:
I agree about the configuration file! It will be done at some point and it will be also useful to fix the crash of CrazyCat00's English translation for KH2FM, since the language auto-detection can not detect the translation programmatically.

Anyway, returning in topic, I reverted the breaking change to load game files from a different folder. So the executable will load game files as before, plus the folder can be specified via arguments.

PR #151, opened 2020-06-13
Doc: Fix GitHub link navigation

Delta-473:
A lot of links didn't have .md behind them so navigation didn't work in GitHub.
I think I have changed all of them.


Vladabdf:
The reason for that is it would destroy browsing on [our website](https://openkh.dev/). While the thought is appreciated, the majority of current documentation is preferred to be browsed there instead of directly from the repository, so the links will remain as is.

Xeeynamo:
I am going to re-open this, since it is not going to break the website.
For example, the [homepage](https://github.com/Xeeynamo/OpenKh/blob/master/docs/index.md) uses links that ends with `.md` and it still work. The reason is because the page builder is smart enough to detect `.md` and converts it into `.htm`.

Delta-473:
Yeah I did notice that the page builder did work correctly because there were already links with .md and they worked. If there's a link broken it could be that I made a typo I haven't tested all the link changes. :)

PR #152, opened 2020-06-14
OpenKh.Game: Allow gamepad inputs

Rikux3:
This PR should enable the usage of a gamepad. 

- Tested with a wired DualShock 4 Controller (without additional software!)
    - Tests with other gamepads (Xbox, others) would be appreciated
- Gamepad and Keyboard inputs work at the same time
- If you disconnect and reconnect the gamepad while the game is running, the controller will be recognized again

One thing to note: The `GamePad` class holds the information `IsConnected`. In my testings it always returned `false` when using the DS4. This seems to be a driver issue.


Rikux3:
I think that should do the trick. 
While searching on how to handle deadzones in Monogame in general i've stumbled upon [this issue.](https://github.com/MonoGame/MonoGame/issues/7046) When you call `GamePad.GetState()`, you can pass an enum on how the handle the deadzones. But the actual values are hard-coded.

Xeeynamo:
Ok, I finally had time to test it. Everything looks fine but the input for the title screen. Do you know what might be the cause of the issue?

Xeeynamo:
You might want to rebase your pull request on master due to this pull request https://github.com/Xeeynamo/OpenKh/pull/144

Rikux3:
I've updated the codebase a little bit. @Xeeynamo could have a look over it?

PR #155, opened 2020-06-18
Add game configuration

Xeeynamo:
Since it was asked by many of you.

For most of the configuration parameters, they will affect the game in real time as soon as you save the change.

The configuration also adds `enforceInternationalTextEncoding`, which enables compatibility with unofficial KH2FM translations.

config

### DIFF
--- a/OpenKh.Game/Config.cs
+++ b/OpenKh.Game/Config.cs
@@ -111,9 +111,9 @@ namespace OpenKh.Game
 
                 fsWatcher.Changed += (object sender, FileSystemEventArgs e) =>
                 {
-                    Log.Info("Configuration file has been changed");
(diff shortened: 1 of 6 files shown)


Xeeynamo:
I imagine you are running the game by changing the working directory, am I right?
The configuration is indeed read and created in the same directory where the binary files are located, but that is not the case if you change the working directory.
I would just recommend to set the game files directory as an application argument:
image

If you still want to change the working directory AND make the game working, you need to modify the `config.yml` and set `dataPath: .`

Please let me know if you still have issues even with my suggestions..

Rikux3:
The working directory was the culprit. When passing as a parameter, everything works perfectly.
~~Now how do i approve this again? :D~~

PR #158, opened 2020-06-28
15jigsaw.bin, jiminy.bar and mixdata.bar structures and documentation

Rikux3:
This PR got a little out of hand because i've only wanted to look at 15jigsaw.bin. But here we are :D
I've omitted editors for these files for now.


Xeeynamo:
I rebased this pull request on master by doing `git rebase origin/master`. It should not have changed anything special, but I ask you to double-check and leave the merging to you.

Since we moved `nQuant.Core` as a sub-module, you will need to launch a `git submodule update` to build the solution successfully.

Xeeynamo:
Actually, before merging as Delta-473 pointed to this Pull Request, is that navigation through GitHub is broken. As he mentioned in #151, if you specify the path of a file without the `.md` extension (eg. check your `jiminy.md`), GItHub would throw a 404. It does not affect the website, but it's better to keep everything working.

PR #163, opened 2020-07-17
Add conditional loader parser for KH3

Xeeynamo:


Xeeynamo:
it misses a proper tool and documentation, then I will convert the draft as a proper pull request. Meanwhile, be free to check and use the parser.

PR #173, opened 2020-07-21
Game title menu

Xeeynamo:
Very basic, but accurate implementation of the "New Game" menu in the title screen.

This aims to define the fundamentals of in-game menu creation and the extension of the title screen.


Xeeynamo:
KH2 JP
image

KH2 IT
image

KH2 FM
image

KH2 PS4
image

The text for japanese builds is un-aligned. I wonder if there is a clean way to adjust it. I doubt that we have to put if-then-else everywhere based on the language used.

Rikux3:
KH2 DE
grafik

KH2 FR
grafik

PR #180, opened 2020-07-24
Update readme

Vladabdf:
Added an image to show the user which runtimes they may need to install, and changed the download link. The desktop runtime we were linking to wasn't working for anyone.


Xeeynamo:
I would specify that both CLI tools and the game engine are cross platforms.
I did not tested it yet, but I believe that the Layout Editor runs on Linux and MAC OS as well.

Vladabdf:
Added a disclaimer that they should work across other systems.

PR #184, opened 2020-07-25
Fix various major issues with Layout Editor

Xeeynamo:
- [x] "Show at frame" input box was writing the wrong value.
- [x] Timeline items can be clickable even when a pop-up window is in front of them.
- [x] Timeline buttons are clicked continuously.
- [x] Unable to assign sprites to a new animation.

### DIFF
--- a/OpenKh.Tools.LayoutEditor/AppLayoutEditor.cs
+++ b/OpenKh.Tools.LayoutEditor/AppLayoutEditor.cs
@@ -232,7 +232,7 @@ namespace OpenKh.Tools.LayoutEditor
                 sequenceProperty.AnimationGroup = sequenceProperty.AnimationGroup;
 
             var frameStart = sequenceProperty.ShowAtFrame;
-            if (ImGui.DragInt($"Show at frame##{index}", ref animGroupIndex))
+            if (ImGui.DragInt($"Show at frame##{index}", ref frameStart))
                 sequenceProperty.ShowAtFrame = frameStart;
 
             var position = new int[] { sequenceProperty.PositionX, sequenceProperty.PositionY };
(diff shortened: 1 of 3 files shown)


Xeeynamo:
This is still not as perfect as I wanted, since it is still possible to click the A, L, P buttons while there is another window in front of the timeline. But it fixes the deletion of those, which was even worse.
Meanwhile, I fixed everything else.

Vladabdf:
So far, two crashes, though one of them is probably unrelated to these updates.
1. Opening ending.2ld and trying to open the sprite group editor caused a crash.
2. When the "Sprite group" value is set above the maximum amount, it crashes (understandably).
Image example: camp.2ld, there are only 134 sprite groups, so increasing it by one causes it to crash.
image

Other than that, the other fixes seem really good!

Xeeynamo:
> So far, two crashes, though one of them is probably unrelated to these updates.
> 
> 1. Opening ending.2ld and trying to open the sprite group editor caused a crash.
> 2. When the "Sprite group" value is set above the maximum amount, it crashes (understandably).
>    Image example: camp.2ld, there are only 134 sprite groups, so increasing it by one causes it to crash.
>    image
> 
> Other than that, the other fixes seem really good!

You're perfectly right. Along with this bug, I discovered even more bugs that could cause crashes in the Layout editing screen. I hope that this will fix every important bug discovered so far 👍

Vladabdf:
Sounds good. Your time and effort on this editor has been invaluable and it's easily one of the most advanced tools for the game out there. Keep up the great work and I'll continue giving feedback as usual! 👍

Xeeynamo:
Anyway the PR is ready to be reviewed

PR #185, opened 2020-07-25
Implement exports in Kh2SystemEditor

kenjiuno:
2020-07-25_22h17_31

2020-07-25_21h59_31


kenjiuno:
I have only KH2 Japanese and KH2FM, thus I cannot verify whether text output is good or not. Please someone test it.

2020-07-25_21h49_32

2020-07-25_22h25_18

kenjiuno:
I guessed that UCM (universal character modifier) discussed at discord may be associated with `Id` of item table. Thus exporting item list should help knowing about ARD file structure. Just I guessed.

kenjiuno:
Copy encoding idea from Place Editor:

2020-07-26_22h11_00

Export seems to be ok too:

2020-07-26_22h09_33

kenjiuno:
Thanks for reviewing! I'll check it

kenjiuno:
OK this will be better...

Xeeynamo:
I did not know you had to duplicate the entire class to allow to export them as any format. I am sure that later on a common solution will be found and we can refactor it. For now, I will be happy to merge the changes!

PR #186, opened 2020-07-25
Fix palette id shift on save, and implement hex color edit.

### LINKED ISSUE #174: Kh2SystemEditor: saving without changes modifes the file
From the looks of it, something in FTST is shifted by 1.
Load 2FM 03system.bin, and directly save it to reproduce.

grafik

[KH2.zip](https://github.com/Xeeynamo/OpenKh/files/4959192/KH2.zip)

kenjiuno:
This is very sober method while it is the best way to enter precise value.

2020-07-25_23h29_22

Fix #174


kenjiuno:
I'm unaware about this color palette meaning. So I have implemented minimal editor so that it can observe the change result.

There is color picker common dialog equipped in Windows. We can summon this.

image

However this lacks alpha channel.

We can ask it later. But it is not smart way.

image

```vb
?InputBox("Oh it is missing alpha value! Please enter new value belo:", , "128")
```

So I have chosen a method to ask for users to enter a hex value `AARRGGBB`.

Xeeynamo:
Hi, I can confirm that the bug is definitively fixed 🎉 . 

Regarding the new way of picking the colour, I like the second idea more: having the ColorDialog from Windows and then an input box that asks for the alpha value. And probably it would also be a good idea to specify that `80` is 100% of opacity.
With WPF, if you bind a `byte` to the input box, the value will automatically be validated to be between `0` and `255`, so there is no need to add any custom validation.

kenjiuno:
method 1: input hex value

method1

method 2: ColorDialog + alpha value input (use binding)

method2

method 3: WIP

wpftoolkit https://github.com/xceedsoftware/wpftoolkit/wiki/ColorPicker

like the following pick implemented in LayoutEditor:

2020-07-26_21h15_49

kenjiuno:
method 3: ColorPicker https://github.com/xceedsoftware/wpftoolkit/wiki/ColorPicker

method3

kenjiuno:
Thank you for checking! I'll merge this now.

PR #189, opened 2020-07-27
Add simple file mod support

Xeeynamo:
This is a quick one. I often found that when I wanted to test modded asset on OpenKH game engine, I had to modify the original game files and often accidentally losing the source.
This adds mod support. By default it tries to find files in the `mod` folder. If the file is found, it loads the modded one. If it is not, it is going to fall back to the original file.

Of course you can set any path from the `config.yml`. I personally set the VFS of CC's DynamicLoader.


Vladabdf:
Excited to test this when I get home!

PR #191, opened 2020-08-02
Add editing actions to ImageViewer

kenjiuno:
Add editing actions:

2020-08-12_15h03_11


kenjiuno:
This pull request is under development.
If someone feels this useful, please share your ideas!

Xeeynamo:
The features you're adding are very interesting and we needed similar logic to this tool :) 
Are you going to put those actions inside a sub-menu later on? (eg. `Edit` menu)

Vladabdf:
For what it's worth, I would prefer these buttons outside of the menu in this instance. Having them easily presented and accessible for such a simple tool is ideal in my opinion. Otherwise, so far it looks like a great addition to have!

kenjiuno:
This time I have put editors under Edit menu.

2020-08-03_09h11_25

In the future, it will be better if it can implement thumbnail list (and action menu too) like the following design sample:

2020-08-03_09h21_41

kenjiuno:
This will be better

kenjiuno:
Thanks for reviewing this!
I'll try to fix it by using assembly reference.

PR #204, opened 2020-08-05
Create dependabot.yml

Xeeynamo:
This will allow to automatically upgrade NuGet packages


Xeeynamo:
@kenjiuno It is quite easy to set-up: on any public repository you just go to "Insights" tab, then "Dependency Graph" and finally select "Dependabot!

PR #214, opened 2020-08-08
Kingdom Hearts II Map Studio

Xeeynamo:
Adds a tool called Map Studio, which is responsible to modify different aspects of information from MAP and ARD files. It also serves as a viewer.

image

Planned functionalities:
- [X] Free camera movement
- [X] BOB editing
- [ ] Verify if BOBs can be edited by attaching to PCSX2
- [x] View collisions
- [ ] Edit collision properties (flags, etc.)
- [x] Switch to different COCT files
- [ ] Handle DOCT files
- [x] Export collision layer
- [ ] Import collision layer
- [ ] Export MAP meshes
- [ ] Import MAP meshes
- [ ] UVSC implementation
- [ ] TEXA implementation
- [x] Enumerate SpawnPoint files
- [x] Allow to modify SpawnPoint objects
- [x] Show and edit event collisions
- [ ] Show and edit walk paths
- [x] Show decompiled SpawnScript
- [x] Allow to compile SpawnScript
- [ ] Simulate SpawnScript


Xeeynamo:
Marked the pull request as Ready. I know I implemented half of the functionality I actually wanted, but many of them depends from aspects of the game that has not been reversed engineered yet. A second pull request can always be opened in the future to add the "missing" functionalities.

Delta-473:
The ability to change the key bindings would be handy for people who don't have qwerty layout.

Tomk1004:
maybe allow the option to move the camera up or down with "Q" and "E" respectively, this saves time having to turn the camera upwards and move up that way

Disgustor:
Shift should increase the speed by a lot more, and highlighting a selected object + targeting the camera on that object would be a much appreciated feature. As it is, there's no way to tell which instance of PLAYER.mdlx you want to edit without moving every instance in the map until the one you want to edit moves. Same goes for bounding boxes.

Xeeynamo:
> I'm sorry it's taken me this long to write the review, anyway, I'd say it's all good to be integrated with the main branch, all the main functions work perfectly fine, and any changes can simply be pushed in future iterations

Thank you so much for your feedback Tom!

Rikux3 (approved):
I'll approve this then.

PR #216, opened 2020-08-12
Decode FooterData of model texture

kenjiuno:
Decode FooterData of model texture.
`OpenKh.Research.MapTexFooter` command line tool is a sample usage of `TextureFooterData`.

```cs
                var modelTexture = ModelTexture.Read(entry.Stream);

                var footerData = modelTexture.TextureFooterData;

                Console.WriteLine(
                    new SerializerBuilder()
                        // omit unimportant and massive array elements.
                        .WithAttributeOverride<TextureAnimation>(it => it.SpriteImage, new YamlIgnoreAttribute())
                        .WithAttributeOverride<TextureAnimation>(it => it.SlotTable, new YamlIgnoreAttribute())
                        .Build()
                        .Serialize(footerData)
                );
```

Doc is ready:
- https://github.com/kenjiuno/OpenKh/blob/tex-footer/docs/tool/CLI.TexFooter/index.md


Xeeynamo:
Excited for this PR :)

Xeeynamo:
With very little effort, this is what I was able to achieve thanks to your parser:
texscroll

I am still reviewing, but so far it's very exciting!

OpenKh.Command.TexFooter/Subcommands/ExportCommand.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        [Required]
+        [FileExists]
+        [Argument(0, Description = "Map file")]
+        public string MapFile { get; set; }
+
+        [Argument(1, Description = "Output dir")]
+        public string OutputDir { get; set; }
+
+        protected int OnExecute(CommandLineApplication app) => OnExecute();
```
Xeeynamo, line 31: Unfortunately having a second `OnExecute` makes the application crashing

OpenKh.Command.TexFooter/Subcommands/ExportCommand.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            var list = new List<string>();
+
+            var barEntries = File.OpenRead(MapFile).Using(Bar.Read);
+            var entry = barEntries.FirstOrDefault(entry => entry.Type == Bar.EntryType.ModelTexture);
+            if (entry == null)
+            {
+                return 1;
+            }
+            if (entry.Stream.Length < 0x60)
```
Xeeynamo, line 43: Maybe a good idea would be to use `ModelTexture.IsValid(stream)` instead?

Xeeynamo (changes requested):
I have few request changes to suggest before approving it:

* The `TextureFooterData` is not tested. There is a `ShouldWriteBackTheExactSameFile`, but the decorator to run the unit test is commented. Would you be able to use a custom texture (to not upload copyright stuff) that will be used for that test?
* The tool Command.TexFooter crashes with the command `export`. I pointed where the problem is.
* Exporting the footer from maps works, but it only works for the first texture it finds. Some maps have `MAP`, `SK0` and `SK1`, but the tool deserialize only the first one that is found.

A small note for the future: do you think would be a good idea to always deserialize the footer in `ModelTexture` object? Because in theory we always want to read the footer after reading the image. It's not a require change but it might be something to consider in the future.

OpenKh.Command.TexFooter/Subcommands/ExportCommand.cs
kenjiuno: Thanks, fixed!

OpenKh.Command.TexFooter/Subcommands/ExportCommand.cs
kenjiuno: Thanks, `IsValid` added!

kenjiuno:
Sorry I have missed comments on this pull request. I have updated the corresponding codes.

Xeeynamo (approved):

PR #224, opened 2020-08-16
COCT refactoring

Xeeynamo:
This PR integrates the idea of the `CoctLogic` class into the `Coct` itself, simplifying the creation of COCT files and hiding a good amount of logic. I am 100% sure this is generating COCT complaint files thanks to the commented test `CollisionTests.PreserveCollision`, but it breaks compatibility with LightCollision, since they have a weird value as `BoundingBoxIndex`. This specific part will be fixed but it is not the scope of this PR.

This is just the first part of the COCT refactoring I have in mind, but I am proposing the change now because I do not want to go too much in depth with the changes yet, in favour of continuous integration.

A bit off topic, but there is another branch called `feature/collision-optimization-wip` that refactors the `VertexIndex` part as well. But even if it was generating virtually the same files, the output contained different indices when compared to a COCT generated by SquareEnix. They are not read in order as I found a `0 1 2 3 8 9 10 11 12 13 6 7 4 5 14 15` pattern on how vertices are stored in the file. Probably they are performing some logic that we are missing at the moment.


OpenKh.Kh2/CoctLogical.cs
```diff
@@ -203,9 +142,9 @@ public class CoctCollision
             public short Vertex2 { get; set; }
             public short Vertex3 { get; set; }
             public short Vertex4 { get; set; }
-            public short Co5Index { get; set; }
-            public short Co6Index { get; set; }
-            public short Co7Index { get; set; }
+            public Plane Plane { get; set; }
```
kenjiuno, line 145: `CoctLogical` and its independent child-classes are for mainly WPF MVVM binding support purpose.
`CoctPlane` is class (object), and `Plane` is struct (ValueType).
ValueType may not be suitable for WPF binding usage, because the writing to properties (like X, Y, Z) will not be reflected.
If we discontinue _OpenKh.Tools.Kh2MapCollisionEditor_ (using WPF) in near future, this is possible decision.

Xeeynamo: Kh2MapCollisionEditor will be discontinued at some point. Right now it might still be useful for debugging purposes, so I will restore it as you suggested 👍

Xeeynamo:
Kh2MapCollisionEditor fixed. Thanks for the feedback :) be free to accept it if you think it is correct.
Before merging though, I would like to have people like @Disgustor or @Tomk1004 to test it, in case of any regression.

Xeeynamo:
I converted this pull request from Ready to Draft [due to a problem](https://discordapp.com/channels/409140906625728532/575294743852875787/745047864295555162) that @Disgustor discovered.


The problem seems to be related to the following commit: 7aa2ace597f0ec15b2132e1605267222e23aa319 on `CreateDoctFromCoct()`. This PR can not be merged into master until this problem is fixed.

PR #226, opened 2020-08-18
Add MapGenTest.

kenjiuno:
- Move MapGen implementation to MapGenUtil (for use with unit test).
- Add `OpenKh.Command.DoctChanger.exe dump-doct` command (for use with unit test).
- Add 4 fbx files for mapgen test.
- Add MapGenTest.


Xeeynamo (approved):
This is a very good idea 👍 I also created another pull request to refactor this by a bit

PR #228, opened 2020-08-18
Fix sequence interpolation

Xeeynamo:
I found we were using the wrong interpolation. It is not cubic, but it is an ease in-out interpolation, which have a slow start and end but a fast middle.

Before the fix:
image

After the fix:
image

---

Notice here how the fixed version fades out the "light" on the title screen quicker and the animations of NEW GAME and LOAD are way more accurate now:

Before the fix:
unfix_1

After the fix:
fix_1

---

I noticed that the interpolation was wrong while I was developing the menu. I clearly remembered that the animation when you open the menu slows down when close to the end of the animation. This was definitively not happening before the fix, as the menu would instantly appear.

Before the fix:
unfix_2

After the fix:
fix_2

### DIFF
--- a/OpenKh.Engine/Renders/SequenceRenderer.cs
+++ b/OpenKh.Engine/Renders/SequenceRenderer.cs
@@ -137,7 +137,7 @@ namespace OpenKh.Engine.Renderers
 
         private void DrawAnimation(Context contextParent, Sequence.Animation animation, int index)
         {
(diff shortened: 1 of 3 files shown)


Xeeynamo:
Tests are now green. They were broken because, after the interpolation algorithm has changed, as consequence some expected values were not equal to the actual ones anymore.

Rikux3 (approved):
Looks good!

PR #229, opened 2020-08-18
Test refactoring

Xeeynamo:
Create dedicated Test projects for each macro category in the project.
This PR will suggest the following structure:
 - `OpenKh.Tests` for low level libraries
 - `OpenKh.Tests.Commands` to test command line tools
 - `OpenKh.Tests.Engine` to test the engine (used by tools and game) or the game engine itself

By following this structure, in future we can also add `OpenKh.Tests.Tools`, but right now there is no need to do so.


kenjiuno:
I haved tested test on 4f1df095384a4a0696d411d2b95fe283b0338dc5,

Tests seems to be fine except on MapGen COCT/DOCT dump compare.

2020-08-20_19h32_29

There is LF/CRLF problem on Windows development... I have an idea (try normalization with `.Replace("\r\n", "\n")`) but I think it is not permanent solution. Is there good idea for this issue...?

2020-08-20_19h31_46

Xeeynamo:
Oh this odd, since both my Windows OS and the Azure Build Agent marks the test as green.
I noticed that xUnit provides a `ignoreLineEndingDifferences` parameter, which is set to `false` by default. I pushed a change that sets it to `true`, so please let me know if the test now passes on your development environment.

kenjiuno:
> I noticed that xUnit provides a `ignoreLineEndingDifferences` parameter, which is set to `false` by default. I pushed a change that sets it to `true`, so please let me know if the test now passes on your development environment.

Thanks, the all tests passed!

I noticed `ignoreLineEndingDifferences` this time... I'll take it in later case!

kenjiuno (approved):

PR #241, opened 2020-09-03
Add menu engine and replicates KH2 menu

Xeeynamo:
This is another big pull request and I am sorry about that. There was a lot of stuff I learned while writing the code for this pull request and I had to change and optimise some aspects of the code bit by bit.

The big thing about this is the implementation of the in-game menu, which is fully animated:
image

There are only two options implemented:

1. Config: all the options are here. There is no scrollbar and the scrolling is a bit clunky. But all the options are there. There is no persistence in those settings as we miss the save functionality.
image

2. Debug menu: While pressing R2 (or the I on your keyboard), the Config entry will return into DEBUG, to access the same functionalities of the terrible debug screen we already have. Adding elements is very easy here.
image

3. Template menu: For all the unimplemented options, a template menu is invoked. It is just a showcase on how a menu can be created and it can be useful to use as a quick reference. It is basically used as a place-holder.
image

I tried my best putting all the common logic in one class: `MenuBase`. That is responsible to handle the open and close animation, on top of stacking sub-menu.

As knowledge sharing, something I want to specify on this implementation is that the menu is re-created every time you change an option. It is not ideal, but it is very, very complicated to have a nice solution that is code-efficient. I was inspired by the retail game, which apparently re-creates the menu as well every time you change option with Up and Down.


Vladabdf (changes requested):
The menu works great! There is one issue that should be addressed ASAP though, as it causes a soft-lock. If you open a handful of fake/template sub-menus and then press Start to exit quickly, and then reopen the pause menu, you cannot enter sub-menus anymore and you cannot close, forcing the user to restart Openkh.Game. If this can be fixed, I will approve the PR gracefully.
A second, unimportant issue is upon entering said fake/template sub-menus, the floating orb animation resets every time while every other animation does not. It's inaccurate, but a minor detail at this time.

Xeeynamo:
Thank you for testing it so quickly. Effectively I did not tested what happens when you are in a deep sub-menu and you press the Start button. I added few fixes including the one you suggested.

About the lum that is resetting itself, unfortunately it is very tricky at the moment to find a fix for it. The current use of it is very different from how the retail game uses it. In the in-game implementation, the lum is always on top of the HUD. Plus when you change option in the menu, the lum is supposed to move smoothly, but in our implementation is just suddenly change position. It will stay inaccurate for a while, until someone will find a solution for it.
image

Xeeynamo:
This might be almost un-noticeable, but the text casts a shadow in the menu in the original game. I was able to replicate it.

image

Vladabdf (approved):
I finally re-reviewed this and it works great for how much effort you put into it. I'm amazed our menu has gotten so far, so I'm gonna go ahead and merge. For now, the main thing we need in it is the debug menu, which works phenomenally. Accepted!

PR #242, opened 2020-09-05
Layout Editor improvements

Xeeynamo:
This is based from [@Vladabdf suggestions](https://discordapp.com/channels/409140906625728532/575294743852875787/751213000538456204). I do not know if I will be able to implement everything in a reasonable amount of time, but most of them are definitively good quality improvements.

A big note to the "Attach text": that is not actually only for text, but also for animations. Please look the commit messages to know which changes will be performed during the development of this PR.


Xeeynamo:
I think there is no reason to hold this back anymore. Be free to review it.

Vladabdf (approved):

PR #247, opened 2020-09-27
Add some basic documentation for KH 1, DDD and ReCoM

Delta-473:


Vladabdf (approved):
Good foundation. I added some info that I had readily on hand, such as release dates, a few descriptors, and checksums. Although, the DDD checksums I added seem incorrect so we'll re-validate those at a later time. For now, thanks for going through and making these documents to begin with!

Issue #223, opened 2020-08-16
4-bpp IMGD issue


kenjiuno:
- _Windows 4-bpp bitmap_ and _PS2 4-bpp bitmap_ share same pixel order. No conversion is required: `0x12 0x34 0x56` is interpreted as `1 2 3 4 5 6`.
- However only **4-bpp unswizzled IMGD** is swapped: `0x12 0x34 0x56` is `2 1 4 3 6 5`

PR #253, opened 2020-10-14
Fix problem of loading KH2 Map radar images in ImageViewer

### LINKED ISSUE #243: ImageViewer fails to open KH2 Map radar images
### Repro Steps
1. Use BarEditor to export the `rada` section from any `.map` file
2. Try and open the exported section in ImageViewer
3. Crash.

Some initial debugging shows the code isn't reading the clut from the file. The `rada` sections appear to be fully valid TIM2 blobs but the picture header has a clut size of 0. The color count is set correctly which suggests an obvious fix.

kenjiuno:
For testing:

Open al00.map with BAR editor, select `rada`, click export, save as `rada.tm2`, and open it with ImageViewer.

2020-10-14_09h35_33

2020-10-14_08h51_04

The pixel swapping problem in 4-bpp images should be handled by #223

Fix #243


Vladabdf (approved):
Works as expected. Thanks for the fix!

PR #260, opened 2020-10-27
Specify line ending and formatting rules

Xeeynamo:
I need to gain again some momentum in this project. This is a very small PR that suggests contributors to strict code formatting rules. I also became aware that shell scripts on Linux do not work with `crlf` line ending. Since there are minor inconsistencies between the text encoding and line ending, the rules will force the editor to always use UTF-8 without BOM and `lf` in all occasions with few exceptions.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,56 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+tab_size = 4
+
+[*.md]
+end_of_line = crlf
+
+[*.sln]
+end_of_line = crlf
+
+[*.csproj]
+end_of_line = crlf
+
+[*.cs]
+end_of_line = crlf
+csharp_new_line_before_open_brace = all
+csharp_new_line_before_else = true
+csharp_new_line_before_catch = true
+csharp_new_line_before_finally = true
+csharp_new_line_between_query_expression_clauses = true
+csharp_indent_case_contents = true
+csharp_indent_switch_labels = true
+csharp_indent_labels = one_less_than_current
+csharp_indent_block_contents = true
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,56 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = space
```
kenjiuno, line 8: Perhaps this will be `indent_size = 4`
According to https://editorconfig.org/

.editorconfig
```diff
@@ -0,0 +1,56 @@
...
+tab_size = 4
```
kenjiuno, line 9: `tab_size = 4` → `indent_size = 4`

kenjiuno (commented):
Thanks, please check comments about `.editorconfig`

I also recommend to run `dotnet-format --workspace OpenKh.sln` once.

```bat
dotnet tool install -g dotnet-format
dotnet-format --workspace OpenKh.sln
```

This will enforce your changes immediately, and then for incoming pull requests, it will reduce extra diffs coming from white space changes.

.editorconfig
Xeeynamo: Sorry, I did not paid attention to it. Fixed 👍

.editorconfig
Xeeynamo: I suggest to keep `tab_size` for the code that already use tabulations rather than spaces

Xeeynamo:
> I also recommend to run `dotnet-format --workspace OpenKh.sln` once.

I was not aware of that tool. Thank you! Although the problem is that I get a lot of files changed. I prefer to not push it as it would give big problems when merging pull requests.

kenjiuno (approved):

PR #261, opened 2020-10-27
Added dictionaries for multiple IDs

osdanova:
Added lists for commands, enemies, equipment, events, inventory, objs and shops's IDs for reference in other files.


Xeeynamo (approved):
Might be very useful 👍 gotta merge it now. Thanks for your contribution!

PR #270, opened 2020-11-08
03System update

osdanova:
Ordered the subfiles, added some missing information, including the whole shop  subfile,  and added the structure section on subfiles that didn't have them.


Xeeynamo (approved):
Very nice 👍

PR #275, opened 2020-11-12
[WIP] Add basic documentation for side games

Delta-473:
This is an early PR just to have feedback on what info to include, layout etc. Atm only the build list is planned for the other games as I don't know much about the file types the Nintendo DS uses. If you have a docs about it feel free to just add them to the PR.
For the DS games I'm gonna do 1 index.md file with links to common info and game specific info because those 2 games are so similar internally.


Xeeynamo:
Excited for this one :)

Delta-473:
I didn't have time to look into the file types for the DS games. So it's mostly builds, some link fixes and basic layout.

Xeeynamo (approved):
In overall I would say it is good. There are few pages where there is almost no content, but it's fine by me as it gives the ground to update those documents later on. I am merging it 👍

PR #278, opened 2020-11-12
ImageViewer: clear current bar entry data before saving

kenjiuno:
Fix OpenKh.Tools.ImageViewer doesn't clear bar data before saving image, when it is launched by OpenKh.Tools.BarEditor as embedded tool.


kenjiuno:
This is repro and explanation of the problem:

2020-11-13_00h19_27

Xeeynamo (approved):
Ooooh this was an interesting bug I overlooked. Thank you for your great contribution Kenjiuno-senpai 🙏

PR #286, opened 2020-11-16
Fix 285: out of bound Spawn points bug

### LINKED ISSUE #285: OpenKh.Game: Rabbits House and The Spooky Cave spawn point out of bound crash
On KH2 EU English version those 2 maps crash the engine because there are -1 spawn points on the map.

afbeelding

Delta-473:
fix #285


Xeeynamo (approved):
Thank you for the fix :)

PR #292, opened 2020-11-19
New documentation.

OrcusDorkus:


Xeeynamo (changes requested):
Thank you for your pull request, it looks really good. I have few changes to suggest though. The first one is to separate the header with their content. I know it's such a stupid thing, but our website theme generator doesn't like it.

This:
```
## SUBMENU ON/OFF >>
* Allows docking of submenus in the top-level menu.
```

should be
```
## SUBMENU ON/OFF >>

* Allows docking of submenus in the top-level menu.
```

docs/remasters/15plus25/demo_debug.md
```diff
@@ -0,0 +1,99 @@
+# [Kingdom Hearts 2.5 HD Remix](remasters/15plus25/index.md) - KH2FM+ Debug Menu
```
Xeeynamo, line 1: `(remasters/15plus25/index.md)` should be `(index.md)` . In the way you wrote it, the link would expect to go to https://openkh.dev/remasters/15plus25/remasters/15plus25/index.md. The reason is because the document it's already in `remasters/15plus25/`, so you do not need to specify the full path of `index.md`.

docs/remasters/15plus25/index.md
```diff
@@ -2,4 +2,7 @@
 
 ## General documentation
 
-* [Editions and updates](updates.md)
\ No newline at end of file
+* [Editions and updates](updates.md)
+## Kingdom Hearts 2.5 HD ReMIX
```
Xeeynamo, line 6: Same here: we need a space between that text and the beginning of the next paragraph

Xeeynamo (changes requested):
I add few more comments too

docs/remasters/15plus25/demo_debug.md
OrcusDorkus: Thank you, good catch.

docs/remasters/15plus25/index.md
OrcusDorkus: Anything that improves readability is good to me.

Xeeynamo (approved):
Congratulations for your first contribution in OpenKH 👍

PR #297, opened 2020-11-19
MSET effects doc

osdanova:
MSET effects documentation (And later animations too if somebody knows the structure). I haven't tested the effects myself, they are from Soraiko's doc.

### DIFF
--- a/docs/kh2/file/type/mset.md
+++ b/docs/kh2/file/type/mset.md
@@ -1,0 +1,81 @@
+# [Kingdom Hearts II](../../index.md) - MSET (Moveset)
+
+The MSET files contain animations and the effects related to them (I-frames, hitboxes, effects...)
+It is a [BAR file](bar.md) that contains a list of "slots". Each slot may or may not have a "motion".
+The various actions in the game point to a specific motion by a slot position in this list, thus many of the motions are dummies, since the characters only use some of them.
+Eg: For Sora, slot 0 is idle, slot 4 is walking and slot 604 is the basic attack.
+
+Each motion is a BAR file with 2 entries, the model animation and the effects.
+
+## Animation structure
+
+???
+
+## Effects structure
+
+| Amount | Description |
+|--------|---------------|
+| 1 	 |  Header
+| X 	 |  Type A effect (Happens during X frames)
+| X 	 |  Type B effect (Triggered on a specific frame)
+
+### Effects header
+
+| Offset | Variable Type | Description |
+|--------|---------------|-------------|
+| 0 	 | byte | Type A effect count
+| 1 	 | byte | Type B effect count
+| 2 	 | short | Start offset of the type B effects
+
+### Type A effect
+
+| Offset | Variable Type | Description |
+|--------|---------------|-------------|
+| 0 	 | short | Start frame
+| 2 	 | short | End frame
+| 4 	 | byte | Effect A ID
+| 5 	 | byte | Param size as shorts
+| 6 	 | short[] | Param
+
+### Type B effect
+
+| Offset | Variable Type | Description |
+|--------|---------------|-------------|
+| 0 	 | short | Trigger frame
+| 2 	 | byte | Effect B ID
+| 3 	 | byte | Param size as shorts
+| 4 	 | short[] | Param
+
+## Effect A list
+
+| ID | Effect | Parameter | Size
+|--------|---------------|-------------|-------------|
+| 0 	 | Allows controls (But brings to idle animaiton) | 
+| 1 	 | Allows controls (But blocks the animation) | 
+| 2 	 | Allows controls | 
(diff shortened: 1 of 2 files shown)


docs/kh2/file/type/mset.md
```diff
@@ -0,0 +1,81 @@
+# [Kingdom Hearts II](../../index.md) - MSET (Moveset)
+
+The MSET files contain animations and the effects related to them (I-frames, hitboxes, effects...)
+It is a [BAR file](bar.md) that contains a list of "slots". Each slot may or may not have a "motion".
+The various actions in the game point to a specific motion by a slot position in this list, thus many of the motions are dummies, since the characters only use some of them.
+Eg: For Sora, slot 0 is idle, slot 4 is walking and slot 604 is the basic attack.
+
+Each motion is a BAR file with 2 entries, the model animation and the effects.
```
Xeeynamo, line 8: When you unpack a motion entry in a MSET BAR I can find two files: a small one and a big one. The big one I imagine is the actual motion file that moves a 3D model. The small one (I am looking at a file of 62 bytes long now) is the effect structure you are mentioning here?

docs/kh2/file/type/mset.md
```diff
@@ -0,0 +1,81 @@
...
+| 1 	 | byte | Type B effect count
+| 2 	 | short | Start offset of the type B effects
+
+### Type A effect
+
+| Offset | Variable Type | Description |
+|--------|---------------|-------------|
+| 0 	 | short | Start frame
+| 2 	 | short | End frame
+| 4 	 | byte | Effect A ID
```
Xeeynamo, line 36: Just to polish this doc by a bit: could you add a link to the enumerations below the structures? For example Effect A ID should have a link to #effect-a-list

docs/kh2/file/type/mset.md
```diff
@@ -0,0 +1,81 @@
...
+## Animation structure
+
+???
+
+## Effects structure
+
+| Amount | Description |
+|--------|---------------|
+| 1 	 |  Header
+| X 	 |  Type A effect (Happens during X frames)
```
Xeeynamo, line 19: It would be good to specify this is an array with the size of "Header -> Effect A Count" and that it points to #type-a-effect as a link too. I had to jump back and forward to start to understand what the document was talking about and to figure out the actual length of a field.

Xeeynamo (changes requested):
Hi @osdanova , I did not know that there were "effect" definitions in every MSET entry. This piece of documentation is precious. I will expand it more as I am working on MSET animations. I left some notes on what in my opinion could improve this document a little more. Keep up with the cool stuff ✌️

docs/kh2/file/type/mset.md
```diff
@@ -0,0 +1,81 @@
...
+The MSET files contain animations and the effects related to them (I-frames, hitboxes, effects...)
+It is a [BAR file](bar.md) that contains a list of "slots". Each slot may or may not have a "motion".
+The various actions in the game point to a specific motion by a slot position in this list, thus many of the motions are dummies, since the characters only use some of them.
+Eg: For Sora, slot 0 is idle, slot 4 is walking and slot 604 is the basic attack.
+
+Each motion is a BAR file with 2 entries, the model animation and the effects.
+
+## Animation structure
+
+???
```
kenjiuno, line 12: I have found existing `.anb` documentation by luck. It says about solely Animation structure so it'll be able to link it.

```
## Animation structure

See [ANB](../anb/anb.md) documentation.
```

docs/kh2/file/type/mset.md
osdanova: Yes, that's right. They are always in that order. I changed "animation and effects" to "animation followed by effects" to make it clearer.

docs/kh2/file/type/mset.md
osdanova: Added multiple links for easier navigation.

docs/kh2/file/type/mset.md
osdanova: I changed 1 to "single" and X to " array", hopefully that's more readable?
About the size, 2 thoughts of mine:
1. If you want to know the size, you can just go to the effect/entry and see how long it is. Eg: Type A's last data is an array of shorts in the offset 6, so the length of an A effect is 6 + 2 * (param size)
2. I can't define a specific length, since it depends on the parameter size of each effect.
That being said, I could add another column with the size in bytes if it's preferred (Eg. 6 + 2 * (param size))

docs/kh2/file/type/mset.md
osdanova: Awesome! I didn't see that! I've added the link as suggested. Thanks!

Xeeynamo (approved):
Looks good to me! Thank you for your contribution 🌟 . I look forward at optimising this document more with the information I found this week.

PR #299, opened 2020-11-21
Fix bug where OpenKh.Game always need at least an argument

Xeeynamo:
An user mentioned a [unexpected behaviour](https://discord.com/channels/409140906625728532/567061704849096725/779595641135300608) in the latest builds of OpenKh.Game. I realised that the mistake was that the game was always asking for the game's path as a command line argument. This fixes the issue.

NOTE: From now on you will need to change your debugging configuration by specifying `--path YOUR_GAME_PATH` if you do not want to use the settings from `config.yml`.

### DIFF
--- a/OpenKh.Game/Program.cs
+++ b/OpenKh.Game/Program.cs
@@ -62,8 +62,7 @@ namespace OpenKh.Game
         private static string GetVersion() => ProductVersion;
 
 
-        [Required]
-        [Argument(0, "Content path", "Location of game's data")]
+        [Option("--data <GAME_PATH>", "Location of game's data", CommandOptionType.SingleValue)]
         public string ContentPath { get; }
 
         [Option(CommandOptionType.NoValue, ShortName = "v", LongName = "console", Description = "Show the console output (Windows only)")]


kenjiuno:
I suggest to display a message on unrecoverable exception thrown on startup. So that user can open `openkh.log` file. such as "An error occurred, please check 'openkh.log' file for detail."

The fixed OpenKh.Game (Release build) will throw exception if I don't specify command line, or `config.yml` is still untouched.

2020-12-01_15h14_54

This exception is silently recorded under Application log of Windows event viewer:

2020-12-01_15h14_01

Xeeynamo:
> I suggest to display a message on unrecoverable exception thrown on startup. So that user can open `openkh.log` file. such as "An error occurred, please check 'openkh.log' file for detail."
> 
> The fixed OpenKh.Game (Release build) will throw exception if I don't specify command line, or `config.yml` is still untouched.
> 
> 2020-12-01_15h14_54
> 
> This exception is silently recorded under Application log of Windows event viewer:
> 
> 2020-12-01_15h14_01

I just checked and we are logging that error at the moment:
```
[000.726] ERR A fatal error has occurred. Please attach this log to https://github.com/xeeynamo/openkh/issues
[000.775] ERR Exception: Unable to detect any region for the game. Some files are potentially missing.:
   at OpenKh.Game.Infrastructure.Kernel.DetectRegion(IDataContent dataContent) in D:\Repository\openkh\OpenKh.Game\Infrastructure\Kernel.cs:line 158
   at OpenKh.Game.Infrastructure.Kernel..ctor(IDataContent dataContent) in D:\Repository\openkh\OpenKh.Game\Infrastructure\Kernel.cs:line 66
   at OpenKh.Game.OpenKhGame..ctor(OpenKhGameStartup startup) in D:\Repository\openkh\OpenKh.Game\OpenKhGame.cs:line 95
   at OpenKh.Game.Program.OnExecute() in D:\Repository\openkh\OpenKh.Game\Program.cs:line 105
```

Maybe the reason you can not see it is because we are not catching any exception while debugging. This is because if we catch the exceptions on debug, we can not see where the error is coming from.

`Program.cs`:
image

kenjiuno:
Yes, the log is surely logged.

image

In the Release build, the exception is re-thrown.
I concern that user cannot notice the existence of log file without prompting the kind message like `please check openkh.log`.

Xeeynamo:
We should have a documentation from `OpenKh.Game` at some point, describing also to look `openkh.log` for troubleshooting. We can not use anything like `MessageBox.Show` because Game runs on Windows, macOS and Linux.

Do you have any suggestion?

kenjiuno:
> We can not use anything like MessageBox.Show because Game runs on Windows, macOS and Linux.

Sorry I missed about this. I forgot that OpenKh.Game should run on OSes other than Windows.

kenjiuno (approved):

PR #309, opened 2020-12-03
Optimize 3D model loading and animations

Xeeynamo:
Performs a bunch of optimizations on loading 3D models from Kingdom Hearts II. Using the animated ObjEntry 0x555 (the Shadow heartless) in Twilight Town we were previously obtaining about 50 frames per **minute** (LOL). After this pull request, there is a massive performance boost to about 42 frames per **second** with the same camera position and entities on screen.

There are still different area where I can see room of optimization, but this is a great boost so far.

Changes performed:
* Avoid to re-calculate some immutable information in `MdlxParser` in `07bc8a6`
* Removes `MeshDesc` and use `MeshDescriptor` straight, without re-mapping it in `bd35ab2` and `1bf5c43`
* Removed a lot of redundant code in `MeshLoader`'s class
* Now a model `IModelMotion` have a `ApplyMotion` method to perform an animation (CPU run)


kenjiuno:
Thanks, In Task manager, CPU load is nicely lowered from 22.7% to 4.9%. Finally `MdlxParser` (`IModelMotion` impl) change is good 👍

before:

before

after:

after

kenjiuno (approved):

PR #321, opened 2020-12-15
KH2: MSN file decompilation

Rikux3:
There are still a lot of unknown Opcodes and it misses a compiler/decompiler.


docs/kh2/file/type/msn.md
```diff
@@ -4,34 +4,29 @@ Mission file. They are located in `msn/{language}/` and describe how a certain m
 
 ## Mission
 
-This is the entry point of the file. It is a binary-type file and it always have as a name `{WORLD_ID}{MAP_INDEX}`
+This is the entry point of the file. It is a binary-type file and it's name is always `{WORLD_ID}{MAP_INDEX}`. Example: for the file `EH21_MS101`, where `EH` is the [world](../../worlds.md) and `21` is the map index, the mission name is called `EH21`.
```
Xeeynamo, line 7: Potential typo: `the mission name is called **MS101**`

Rikux3: I guess i wrote that a bit unclear. What i meant is the first subfile of an MSN file always follows the {world}{room} naming scheme.
grafik

1234567890num:
Seems like the bitmask value for "Summon Disabled" is 64 (0x40) instead of 32 (0x20)?

Rikux3:
The documentation doesn't match the source code files right now.

Rikux3:
Although this still misses a compiler/decompiler, i don't want to withhold the informations any longer. It's still incomplete as to how the single opcodes handles the parameters, but i think that can be added over time.

shinerkz:
> 
> 
> Although this still misses a compiler/decompiler, i don't want to withhold the informations any longer. It's still incomplete as to how the single opcodes handles the parameters, but i think that can be added over time.

Deffo can be added over time. Which opcodes are still unknown?

Rikux3:
All opcodes are known by now, but not how the parameters are used. For example, the timer has 3 bytes which i haven't identified yet.

Topaz-Reality:
> All opcodes are known by now, but not how the parameters are used. For example, the timer has 3 bytes which i haven't identified yet.

Didn't CrazyCatz identify them? I remember using a timer feature. Might be worth looking into his tools or asking him.

shinerkz:
> 
> 
> All opcodes are known by now, but not how the parameters are used. For example, the timer has 3 bytes which i haven't identified yet.

Oh, cool. Will double check when I get home, but do you have the bytes for the flashing timer warning and the bytes for when the countdown pops up in the middle of the screen? If not, it might be those.

Xeeynamo:
Is this ready to be reviewed? If yes, could you rebase this upon master first?

Rikux3:
I'm redoing this since i have problems with rebasing this branch.

PR #324, opened 2020-12-19
BBS/DDD PAM animation documentation

akderebur:


docs/bbs/pam.md
```diff
@@ -30,3 +30,73 @@ Animations in Birth by Sleep do not need to be aligned in any way.
 | 06 | byte | Bone count, must match bone count of target model.|
 | 08 | uint16_t | Animation frame count.|
 | 0A | uint16_t | Loop to frame, determines the point the animation loops to.|
+
+## Animation Data
+The animation data consists of separate channels for each animated property. There can be maximum 9 channels for every bone: Translation X/Y/Z, Rotation (radians) X/Y/Z, Scale X/Y/Z.
```
Xeeynamo, line 35: Sorry for being picky: could you add a new line between each header and the paragraph? When you do not do so, it will trigger a `MD022/blanks-around-headings/blanks-around-headers: Headings should be surrounded by blank lines`. You receive that warning when you use VS Code as editor for markdown files.

docs/bbs/pam.md
```diff
@@ -30,3 +30,73 @@ Animations in Birth by Sleep do not need to be aligned in any way.
...
+#### Keyframe Entry
+
+| Variable Type | Description |
+|---------------|-------------|
+| byte   | Frame id of this keyframe (animation frame count <= 255)|
+| ushort | Frame id of this keyframe (animation frame count > 255)|
+| ushort | Quantized value of this keyframe|
+
+## Implementation and Other Uses
+An animation parser/viewer can be made using the documentation above. Although all the information was gathered from BBS, it seems to hold true for DDD. I have implemented this in my [tool for DDD](https://forum.xentax.com/viewtopic.php?p=169270&sid=51dbc614987b03be478bbcb5b90665a4#p169270) remaster on PS4.
```
Xeeynamo, line 100: I would personally prefer to specify _"There is a tool for DDD"_ with the same link, so the documentation would be more generic.

Xeeynamo (changes requested):
Hi and thank you for your first Pull Request for OpenKH! Me and other people are excited about this as no one ever looked in details into BBS/DDD animations.

We do not yet have a PMO parser, but this puts the base to support PAM animations which would help to convert animations across KH1, KH2, BBS and DDD other than natively support their 3D models in the OpenKH game engine. It will be very exciting to be able to crossing assets from different games.

Before I approve this, I left some comments about minor formatting issues.

thamstras:
I have checked this against actual BBS .pam's and it all looks reasonable except for one correction I had to make.
It seems a channel has no data (ie: is constant) when it's keyframe count is 1, not 0 as written.

docs/bbs/pam.md
Tomk1004: Alternately, "There is a tool created by Akderebur"  works, as it both gives you express credit, and also makes the docs more generic

Topaz-Reality:
This is awesome. I do not know what else to say since I am a noob at these things. But it's awesome nonetheless.

docs/bbs/pam.md
Xeeynamo: Good point @Tomk1004 . I agree that yours is a better resolution.

akderebur: I have made the changes and also fixed the error that thamstras pointed out.

Xeeynamo (approved):

Xeeynamo:
Thank you @akderebur 🙏 I am merging it into `master`.

PR #325, opened 2020-12-19
Many additions to KH2TextEditor

Topaz-Reality:
Made many additions, such as:
-> Add a new line when RETURN is pressed.
-> Ability to add strings [Adds a "FAKE"].
-> Ability to remove strings [Revomes selected].
-> A context menu for the TextBox for adding Commands and Icons.
-> A ColorBox for the Color Command.

Demo Images:
image
image
image
image

Known Bugs:
-> Every function that adds to the text causes the Cursor to return to 0. This cannot be negated for now because of the MVVM arrangement.
-> Icon View is broken but that is not because of my changes.
-> Code is a bit ugly, but that can be fixed.

Planned Changes:
-> Fix [fontimage/icon] unswizzling
-> Make a "better" Icons context menu.

IMPORTANT:
The ColorBoxDialog is from https://github.com/drogoganor/ColorPickerWPF which is included as a NuGet package.

### DIFF
--- a/OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
+++ b/OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
@@ -1,4 +1,4 @@
-﻿using OpenKh.Engine.Renders;
+using OpenKh.Engine.Renders;
 using OpenKh.Kh2;
 using OpenKh.Kh2.Messages;
 using OpenKh.Tools.Common.Rendering;
@@ -8,7 +8,12 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Windows;
+using System.Windows.Controls;
 using Xe.Tools;
+using Xe.Tools.Wpf.Commands;
+
+using ColorPickerWPF;
+using System.Windows.Media;
 
 namespace OpenKh.Tools.Kh2TextEditor.ViewModels
 {
@@ -22,6 +27,12 @@ namespace OpenKh.Tools.Kh2TextEditor.ViewModels
         private string _searchTerm;
         private RenderingMessageContext textContext;
         private bool _showErrors;
+
+        public RelayCommand HandleReturn { get; }
+        public RelayCommand HandleAddition { get; }
+        public RelayCommand HandleRemoval { get; }
+        public RelayCommand HandleComm { get; }
+        public RelayCommand HandleIcon { get; }
 
         public List<Msg.Entry> MessageEntries
         {
@@ -56,6 +67,7 @@ namespace OpenKh.Tools.Kh2TextEditor.ViewModels
                 OnPropertyChanged(nameof(Text));
             }
         }
+
 
         public string Text
         {
@@ -127,6 +139,66 @@ namespace OpenKh.Tools.Kh2TextEditor.ViewModels
             Drawing = new SpriteDrawingDirect3D();
             CurrentMessageEncoder = Encoders.InternationalSystem;
             _messages = new MessagesModel(this, this, new Msg.Entry[] { });
+
+            HandleReturn = new RelayCommand(x =>
+            {
+                Text += "{:newline}";                
+                SelectedItem = SelectedItem;
+            });
+
+            HandleRemoval = new RelayCommand(x =>
+            {
+                MessageEntries.RemoveAll(x => x.Id == SelectedItem.Id);
+                ResetMessagesView();
+            });
+
+            HandleAddition = new RelayCommand(x =>
+            {
+                int _id = MessageEntries[MessageEntries.Count - 1].Id;
+                MessageEntries.Add(new Msg.Entry() { Id = _id + 1, Data = new byte[] { 0x33, 0x2E, 0x38, 0x32, 0x00 } });
+                ResetMessagesView();
+            });
+
+            HandleComm = new RelayCommand(x =>
+            {
+                int _o = Convert.ToInt32(x);
+
+                switch (_o)
+                {
+                    case 0:
+                        Text += "{:reset}";
+                        break;
+                    case 1:
+                    {
+                        Color _color;
+                        bool _ok = ColorPickerWindow.ShowDialog(out _color);
+
+                        if (_ok)
+                            Text += "{:color " + string.Format("#{0}{1}{2}{3}", _color.R.ToString("X2"), _color.G.ToString("X2"), _color.B.ToString("X2"), _color.A.ToString("X2")) + "}";
+                    }
+                        break;
+                    case 2:
+                        Text += "{:scale 16}";
+                        break;
+                    case 3:
+                        Text += "{:width 100}";
+                        break;
+                    case 4:
+                        Text += "{:clear}";
+                        break;
+                    case 5:
+                        Text += "{:position 0,0}";
+                        break;
+                }
+
+                SelectedItem = SelectedItem;
+            });
+
+            HandleIcon = new RelayCommand(x =>
+            {
+                Text += "{:icon " + x + "}";
+                SelectedItem = SelectedItem;
+            });
         }
 
         public void SelectMessage(int id) => SelectedItem = Messages.GetMessage(id);
--- a/OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
+++ b/OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
@@ -80,6 +80,7 @@ namespace OpenKh.Tools.Kh2TextEditor.ViewModels
 
                 _currentText = value;
                 selectedItem.Text = value;
+                OnPropertyChanged();
             }
         }
 
@@ -143,7 +144,6 @@ namespace OpenKh.Tools.Kh2TextEditor.ViewModels
             HandleReturn = new RelayCommand(x =>
             {
                 Text += "{:newline}";                
-                SelectedItem = SelectedItem;
             });
 
             HandleRemoval = new RelayCommand(x =>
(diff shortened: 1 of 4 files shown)


OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
```diff
@@ -127,6 +139,66 @@ public TextEditorViewModel()
             Drawing = new SpriteDrawingDirect3D();
             CurrentMessageEncoder = Encoders.InternationalSystem;
             _messages = new MessagesModel(this, this, new Msg.Entry[] { });
+
+            HandleReturn = new RelayCommand(x =>
+            {
+                Text += "{:newline}";                
+                SelectedItem = SelectedItem;
```
Xeeynamo, line 146: This is a bit confusing, even if it works. I have few suggestions to improve it:
On `public string Text` setter, at the very end we can use `OnPropertyChanged()`, so there is no need for a `SelectedItem = SelectedItem`. I also realise that having the `{:newline}` appended at the end regardless the position of the cursor is not ideal...

What do you think if we just get rid of `{:newline}` to replace it with a `\n`? There is an unit test called `DeserializeNewLine` that is responsible to double-check how the new-line is handled. As long as you change it to `Assert.Equal("hello\nworld", commands[0].Text);` and the test is green, the functionality will be implemented.

OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
```diff
@@ -127,6 +139,66 @@ public TextEditorViewModel()
...
+            });
+
+            HandleRemoval = new RelayCommand(x =>
+            {
+                MessageEntries.RemoveAll(x => x.Id == SelectedItem.Id);
+                ResetMessagesView();
```
Xeeynamo, line 152: I do not ask any action here, but this makes me realise how unflexible this tool is. `ResetMessagesView` resets the list and de-select the currently selected item.

What we do in other tools is to use a single list for everything. But this tool uses two separate lists: the effective list of messages and the list shown in the ListBox. 🤮

OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
```diff
@@ -127,6 +139,66 @@ public TextEditorViewModel()
...
+
+            HandleReturn = new RelayCommand(x =>
+            {
+                Text += "{:newline}";                
+                SelectedItem = SelectedItem;
+            });
+
+            HandleRemoval = new RelayCommand(x =>
+            {
+                MessageEntries.RemoveAll(x => x.Id == SelectedItem.Id);
```
Xeeynamo, line 151: Small note: `SelectedItem` can return `null`.

OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
```diff
@@ -127,6 +139,66 @@ public TextEditorViewModel()
...
+                ResetMessagesView();
+            });
+
+            HandleAddition = new RelayCommand(x =>
+            {
+                int _id = MessageEntries[MessageEntries.Count - 1].Id;
```
Xeeynamo, line 157: This assumes that the list is always ordered at that there is at least one element. I would suggest to do the more secure `MessageEntries.Max(x => x.Id) + 1`.

I do not know how much important this is, but IDs between the various MSG files never overlaps, even if they belong to different worlds. I have no idea why the developers took such decision as multiple MSGs are never load. I would not be bothered by this now, but that can be considered in the future.

OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
```diff
@@ -127,6 +139,66 @@ public TextEditorViewModel()
...
+                MessageEntries.Add(new Msg.Entry() { Id = _id + 1, Data = new byte[] { 0x33, 0x2E, 0x38, 0x32, 0x00 } });
```
Xeeynamo, line 158: Here you are assuming that the sequence `0x33, 0x2E, 0x38, 0x32, 0x00` is always `FAKE` regardless the region. I suggest to use `CurrentMessageEncoder.Encode(MsgSerializer.DeserializeText("FAKE").ToList())` to get the byte array. In that way you will be always sure to use the bytes for that specific region encoding. Sorry if it looks a bit verbose.

OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
```diff
@@ -127,6 +139,66 @@ public TextEditorViewModel()
...
+                        break;
+                    case 4:
+                        Text += "{:clear}";
+                        break;
+                    case 5:
+                        Text += "{:position 0,0}";
+                        break;
+                }
+
+                SelectedItem = SelectedItem;
```
Xeeynamo, line 194: Same here for the suggestion I written above

OpenKh.Tools.Kh2TextEditor/Views/TextEditorView.xaml
```diff
@@ -71,9 +77,88 @@
...
 
             <TextBox
+                x:Name="textInputField"
                 Grid.Row="0"
                 MaxLines="7" TextWrapping="Wrap" VerticalScrollBarVisibility="Auto"
-                Text="{Binding Text, UpdateSourceTrigger=PropertyChanged}"/>
+                Text="{Binding Text, UpdateSourceTrigger=PropertyChanged}">
+                <TextBox.ContextMenu>
+                    <ContextMenu>
+                        <MenuItem Header="Icons">
```
Xeeynamo, line 86: The list is really looooong. Can I suggest to split it into categories? Like Icons\Buttons, Materials, Ranks, Weapons, etc. It would allow to reach different elements from the list faster.

OpenKh.Tools.Kh2TextEditor/Views/TextEditorView.xaml.cs
```diff
@@ -20,7 +20,7 @@ namespace OpenKh.Tools.Kh2TextEditor.Views
     /// </summary>
     public partial class TextEditorView : UserControl
     {
-        public TextEditorView()
+       public TextEditorView()
```
Xeeynamo, line 23: Typo?

Xeeynamo (changes requested):
Thanks for this pull request. It is full of quality improvements they should have been there since the very beginning!

I realise that some parts of the code of this tool are clunky. It is my bad as it was a very old tool I written when my skills were just slightly above the acceptable (not that I am a MVVM master now lol). I also realise that one of the most annoying limitations is that for every new item you append it at the very end. I do not know if there is a way to detect the cursor position inside a TextBox, but that would be just a well welcomed change rather than a required one.

A part from everything I left a review comments, the rest looks really good 👍⚡

OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
Topaz-Reality: Well, OnPropertyChanged() changed the string, but not the TextBox for some reason. I had to compromise.
And yes, /n instead of {:newline} would be better imho.

OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
Topaz-Reality: Would be easy to fix. I did not realize this.

OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
Topaz-Reality: Roger. Would be safer to do this as mentioned.

OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
Topaz-Reality: "FAKE" is region independent. That is why. Capital letters currently exist in all regions. Would change if necessary but, I could implement your change regardless.

OpenKh.Tools.Kh2TextEditor/ViewModels/TextEditorViewModel.cs
Topaz-Reality: Same as I mentioned above.

OpenKh.Tools.Kh2TextEditor/Views/TextEditorView.xaml
Topaz-Reality: Could do.

OpenKh.Tools.Kh2TextEditor/Views/TextEditorView.xaml.cs
Topaz-Reality: Nah, VS being stupid.

Vladabdf (approved):
The additions to this editor are wonderful and make it a fair bit more user friendly. In the future the tool should probably be rewritten to allow for a few design changes that even further improve the end user usability, but for now, excellent work. Thanks for your contribution, Topaz!

Topaz-Reality:
> The additions to this editor are wonderful and make it a fair bit more user friendly. In the future the tool should probably be rewritten to allow for a few design changes that even further improve the end user usability, but for now, excellent work. Thanks for your contribution, Topaz!

The tool does need a refactor to it. But I think it can be managed in the future. I will even be the one to do it if possible.
Thanks for your input. I do appreciate the feedback.

Xeeynamo (approved):

Xeeynamo:
I am merging it as it is. The quality improvements are definitively there and the `\n` instead of `{:newline}` can be done later anyway. Nice one @Topaz-Reality 👍

Topaz-Reality:
I appreciate the feedback. Thanks for the review!

PR #312, opened 2020-12-06
Interpolated animation support in-engine

Xeeynamo:
Draft pull request with some ideas on how to implement animations in-game. It is far from being done and I have very little knowledge on how to do it.

The core animation mechanic is found in `Kh2MotionEngine.ApplyInterpolatedMotion`. I copied the code from `MdlxParser.BuildTPoseMatrices` and expanded it as an attempt to try to create an animation. But the pose looks very messy at the moment.

The branch there will be useful to mess around and try to understand how animation works in-game and to replicate them in-engine.

PR #327, opened 2020-12-19
KH2 Event support

Xeeynamo:
Experimental cutscene support for Kingdom Hearts II, also called Event. The format itself is not complex, but the interpreter is. There is still some broken stuff that makes the event playing very inaccurately, but the proof of concept is there:

https://user-images.githubusercontent.com/6128729/102701256-05135480-424d-11eb-8c66-9adc8a4b6ba0.mp4

https://user-images.githubusercontent.com/6128729/102701290-7521da80-424d-11eb-86a3-be6d10dfa119.mp4

There are two main parts of it: `OpenKh.Kh2.Ard.Event` is the event parser itself, that allows to read, write and create events. Then `OpenKh.Game.Events.EventPlayer` is the main engine that reads the parsed event and applies transformation to the scene to play the cutscene. I had to refactor some stuff, such as separating KH2 specific logic that was in `MapState` into `Kh2Field`. This one implements a `IField`, a generic implementation that might be swapped with custom implementations for other games.

I also added a new option in the debug menu that allows to select and play events:
cutscene-selection


Xeeynamo:
I've been working on creating a very accurate (almost identical) camera interpolation Kingdom Hearts II:

Before:
https://user-images.githubusercontent.com/6128729/103043538-0c03d500-4575-11eb-9d8b-b1b3509bd1da.mp4

After:
https://user-images.githubusercontent.com/6128729/103043534-073f2100-4575-11eb-8f5d-5b6f46f6cbd4.mp4

I accurately chosen specific cutscenes from the game and used the camera output and the frame number as test case for our Unit Tests and worked on the algorithm until all of them passed. Interestingly, the interpolation logic is very similar to the one used by model animations, so I suspect we can re-use it for them to accurately replicate the animations too.

I will continue to work on this PR as there is still a lot to do before considering this done.

Xeeynamo:
This pull request is ready to be reviewed. This brings basic cutscene playback such as camera movement, basic subtitle support, actor animations and positioning. The entirety of movie, audio, blur and layout support will come later in separate pull requests to not bloat this one too much.

Small note: If you press L2 during a cutscene you will have a free-cam functionality. If you press R2,you will pause the cutscene that is currently playing, useful for debugging or just to have fun with it.

Rikux3:
hb04 Event 412 crashes the engine. Motion.cs Line 506
grafik

Xeeynamo:
> hb04 Event 412 crashes the engine. Motion.cs Line 506
> grafik

Unfortunately that looks like to be a crash from the motion system merged from #312 . I would fix it but it needs further investigation and it is unrelated to the content of this PR. According to @Delta-473 , that is not the only event that loads an unsupported animation.

Delta-473:
The only 2 crashes I found are: fov being invalid and an out of bound issue on beast castle bridge. Need to make a spreadsheet for which cutscenes work and which ones crashes.

Xeeynamo:
> The only 2 crashes I found are: fov being invalid and an out of bound issue on beast castle bridge. Need to make a spreadsheet for which cutscenes work and which ones crashes.

That would be a very good idea. Do you think we can create a script that cycles through every cutscene of every map in game's data and, with a try-catch, it prints if it crashed and with what message? A simple `while (!_eventPlayer.IsEnd) _eventPlayer.Update(1f / 30f);` should cycle an entire cutscene in few milliseconds.

kenjiuno:
If we write individual batch tests against `.anb`/`.mset` and `.ard`, the xunit private test can be setup, it is ard version   https://github.com/kenjiuno/OpenKh/commit/07507692c42c21ab0f9688f472d1a6defcb51fd9

image

kenjiuno:
`.anb` files seem to be ok.

2020-12-27_15h42_35

`.mset` fails in 91 motions (11,334 passed). I'll check about this in another pr.

2020-12-27_16h05_08

Xeeynamo:
Thanks to the recent rebase on #338 , the correct 3D models are loaded for worlds LK, LM, WI and TR.

Rikux3:
Is this ready to be merged now?

Xeeynamo:
> Is this ready to be merged now?

There are still some problems such as cutscenes that uses animations that leads to a crash. But really those are unrelated to the cutscenes. For what I know this PR does not bring any regression to the existing codebase. There is still some stuff to research but the info I have are currently very incomplete. This is the best I can do for now so IMHO is ready to be merged and improved later on.

Rikux3 (approved):

PR #328, opened 2020-12-20
Fix fontimage swizzling.

Topaz-Reality:
This PR fixes an oversight with RawBitmap swizzling where it applied a fix for 4bpp images to 8bpp images also.

This PR has been tested with everything except for OpenKH.Game and is functional.

Demo Images:
image
image
image

And here are IMGDs that I have tested (Swizzled and non-swizzled):
image
image


Xeeynamo (approved):
Very good stuff 👍 I am merging it.

PR #329, opened 2020-12-20
Fix image 4bpp for real

Xeeynamo:
So, apparently we were using three different portions of code to perform 4bpp image handling, where every one of them had their own implementation. I fixed it by using the single system at `OpenKh.Imaging`. From now on, `image.ToBgra32()` is everything you need to do as long as the image implements `IImageRead`.

The main reason why it was so complicated to get this right is because the problem lied in the PS2 functions `Decode8` and `Encode8` that delivered a different 4bpp bit order than what a PC uses. So every time we were corrupting all the viewers or just the creation process of an IMGD/TM2/ModelTexture.

This should fix everything for good.


Vladabdf (approved):
Tested with the following tools:
Command.ImgTool
Layout Editor
Image Viewer
OpenKh.Game
KH2 Text Editor

Finally, I think all of our imaging woes are solved. Not a single one produced an incorrectly (un)swizzled 4bpp, 8bpp, or 32bpp image, color palettes are correct, and all modified files are readable by KH2 without issue. Thanks for this fix!

Vladabdf:
Upon further inspection, I made an oversight in my testing: Image Viewer currently destroys the alpha channel when replacing images as of this pull request. Everything else functions as desired I believe.

Topaz-Reality:
> Upon further inspection, I made an oversight in my testing: Image Viewer currently destroys the alpha channel when replacing images as of this pull request. Everything else functions as desired I believe.

This issue is confirmed to occur in releases 218 and 219 as well. Has nothing to do with this PR.

Vladabdf:
In that case, I will just make an issue since it appears unrelated to this branch/PR.

PR #332, opened 2020-12-21
A PngImage IImageRead

kenjiuno:
WIP


Xeeynamo:
Hi @kenjiuno , did you consider to use the following NuGet package? It is open source and it does not depend from Windows GDI+: https://github.com/SixLabors/ImageSharp

kenjiuno:
I have checked ImageSharp.

Although ImageSharp will read/write `Indexed4` and `Indexed8` png images, the image is up converted to 32bpp RGBA on load. The color palette seems to be discarded.

image


[pngs.zip](https://github.com/Xeeynamo/OpenKh/files/5725940/pngs.zip)

kenjiuno:
I have not tested this so much though, png image loader is replaced with our own one.

Current `PngImage` is not fully compliant with all kinds of valid png files such as using features like the following ones:

- PNG_INTERLACE_ADAM7
- PNG_COLOR_TYPE_GRAY
- PNG_COLOR_TYPE_GRAY_ALPHA
- 1bpp mono bitmap
- gamma, icc profile and so on

Also code recovery of displaying background checker board.

2020-12-22_06h11_47

Vladabdf (approved):
The result is a much more usable Image Viewer. If there are no objections, I think it's ready to be merged.

Xeeynamo:
Thank you Kenjiuno 🙏 I am merging. If you intend to make it as a NuGet package it will become useful to many people that have the same problem, not only in OpenKH.

PR #333, opened 2020-12-24
Documents  bbs models, maps and collisions

thamstras:
Adds documentation for BBS's models, maps and what's known of the map's collision data.

### DIFF
--- a/docs/bbs/pmo.md
+++ b/docs/bbs/pmo.md
@@ -30,11 +30,11 @@ The Header is followed by `Texture Count` Texture Info records
 | 0x4    | char[0xC] | Texture Name |
 | 0x10   | int32[4] | unknown |
 
-See also: [TIM2]()
+See also: [TIM2](../common/tm2.md)
 
 ## Mesh Sections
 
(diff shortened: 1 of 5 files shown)


docs/bbs/pmo.md
```diff
@@ -0,0 +1,170 @@
...
+| Joint Weights |
+| Texture Coords |
+| Color |
+| Position |
+
+### Reading 'Normalized' values
+
+Vertex properties which are in the format `8-bit normalized` or `16-bit normalized` need to be converted to floating point values before rendering. Note that position values should read as **signed** values, the others as **unsigned** values. The conversion is as follows:
+
+```c++
```
Xeeynamo, line 142: A very small note that will not stop us from merging this: the `c++` is not exactly recognised as the correct one is `cpp`. This is according to Linguist: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L600

Xeeynamo (approved):
I like the quality of information and organisation you put in this document 👍 really appreciated. This will give even a bigger boost to the project once someone will look into implementing the formats you documented here!

PR #335, opened 2020-12-26
Updated more fields and subfiles in the docs.

osdanova:
Patn subfile, Magc fix, ally damage cap on objentry, went and memt subfiles, pref subfile structure.


Xeeynamo (approved):
This will help me as I am discovering some stuff regarding 03system too. Merging it.

PR #336, opened 2020-12-27
Workaround for broken mset motion data

kenjiuno:
Workaround for broken mset motion data that is pointing to overflowed index of keyFrames and transformationValues.

To enable xunit batch test, set `KH2Dir` to valid path to asset files extracted by extractor:

```cs
private static string KH2Dir = @"H:\KH2fm.OpenKh\";
```

and uncomment:

```cs
//[Theory, MemberData(nameof(Source))]
```

And then build and run tests.

image

### DIFF
--- a/OpenKh.Kh2/Motion.cs
+++ b/OpenKh.Kh2/Motion.cs
@@ -1,4 +1,5 @@
 using OpenKh.Common;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


Xeeynamo (changes requested):
This PR gives me some ideas that I will discuss in our Discord server. For now, could you change the path that looks for files from `@"H:\KH2fm.OpenKh\"` to `".tests/kh2_data"`? I did not just come up with this name as it is found [here](".tests/kh2_data").

kenjiuno:
Ok I have changed the asset path to `".tests/kh2_data"`.

Xeeynamo (approved):

PR #337, opened 2020-12-27
Test with all game's data and publish compatibility report

Xeeynamo:
This PR is run unit tests to all the games data and create a report. The report can be downloaded in TRX and HTML. The TRX file can be open in Visual Studio to debug the failed test, while the HTML will be published in the `docs/` folder as an additional pull request that will be automatically created. Only the tests that starts with `Batch_` will be considered by this report tool.

### DIFF
--- a/.github/workflows/advanced-testing.yml
+++ b/.github/workflows/advanced-testing.yml
@@ -1,0 +1,69 @@
+name: Test with game data
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  run-tests:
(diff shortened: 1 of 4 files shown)


Vladabdf:
Closing as I do not understand the importance, relevance, or usefulness of this addition.

Delta-473:
It was not finished but the purpose of it is that you have an automated test that tries to parse every single game file. So that we know which files aren't working correctly.
Eventually you could add also a tests that plays through every cutscene in high speed.
You can shove all the game data in a private docker container and the you can run it as a pipeline action.

PR #341, opened 2020-12-29
03system MEMT EU/US support

Xeeynamo:
In #338 the MEMT supported only the Final Mix version of `03system`, breaking both editor and engine when using the Vanilla files. This adds support to it for the parser, game engine and the SystemEditor tool.

NOTE: I tried this using the EU/US 03system.bin and not the Vanilla Japanese one, so I can not ensure its compatibility.


MarioSonic2987:
Fixes crash in the game engine when using non-Final Mix files (Spanish KH2 in this case):

image
image

Rikux3 (approved):
Tested it with Vanilla KH2 JP, works just like EU/US

Issue #331, opened 2020-12-20
Image Viewer should be able to check if pngquant has been supplied in the root folder


Vladabdf:
The Image Viewer contains an option to use pngquant. Upon enabling the option, the program should scan the folder it's running in for pngquant, and if it is not detected, prompt the user to download their appropriate binary file [here](https://pngquant.org/).

Example of pngquant being enabled but not detected:
![image](https://user-images.githubusercontent.com/20749232/102724638-ef4e6f80-42de-11eb-8ffb-0fa8263639c1.png)

Xeeynamo:
That would be cute to have

Topaz-Reality:
Addressed in Merged PR #351.

Vladabdf:
Thanks for the work. 💯

PR #132, opened 2020-05-31
About Dialog and contributors page

Xeeynamo:
Basics of the proposal to remove my name from the tools and to build a proper contributor page to recognise people's work. This would respect the main concept behind OpenKH project.

The "About" dialog has been duplicated to `OpenKh.Tools.Common` in order to be customised with brand and colours of OpenKH:
image

I also added a draft of `docs/contributors/index.md`, which can give a vague idea on how that page would be structured.

This pull request is a draft. I think that every one of us should give their touch to it and reach something we all agree to.

PR #343, opened 2020-12-30
Add a dialog to inform missing PngQuant on ImageViewer.

Topaz-Reality:
Addresses issue #331. The requested method requires Xe.Tools.Wpf to be edited, with the addition of the files in this archive:
[XE_MessageDialog.zip](https://github.com/Xeeynamo/OpenKh/files/5752958/XE_MessageDialog.zip)

I can also do this with standard WPF DialogBoxes, I think. But since this was the way that was requested, I have done it like so.
Also discovered an issue with Xe.Tools.Wpf. Will be making an issue about the bug.

Preview Image:
image

### DIFF
--- a/OpenKh.Tools.ImageViewer/ViewModels/ImageViewerViewModel.cs
+++ b/OpenKh.Tools.ImageViewer/ViewModels/ImageViewerViewModel.cs
@@ -294,6 +294,24 @@ namespace OpenKh.Tools.ImageViewer.ViewModels
                 x => IsMultipleImageFormat && ImageContainer.Count >= 1
             );
 
+            CheckQuant = new RelayCommand(x =>
+            {
+                if (UsePngquant == true)
+                {
+                    if (!File.Exists("pngquant.exe"))
+                    {
+                        UsePngquant = false;
+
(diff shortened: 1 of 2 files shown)


Xeeynamo (changes requested):
Thanks @Topaz-Reality . I have a proposal though. Rather than modifying the existing library, we could create a copy of that specific dialog so it can even be customised more in the future. My pull request #132 does exactly the same. Actually it's kind-of time that #132 will be continued by someone as I still require input from all of you to de-brand it from my name and put the community at the centre.

Topaz-Reality:
> Thanks @Topaz-Reality . I have a proposal though. Rather than modifying the existing library, we could create a copy of that specific dialog so it can even be customised more in the future. My pull request #132 does exactly the same. Actually it's kind-of time that #132 will be continued by someone as I still require input from all of you to de-brand it from my name and put the community at the centre.

Well, as long as the dialog itself can be copied to other projects, I am down for it!
I will check this out.

Also another issue: Hyperlinks cause a crash when clicked, this is also present in the About Boxes. It seems like this is a Xe.Tools.Wpf issue.

Topaz-Reality:
Planning on remaking this implementation another way. Closing for now.

PR #344, opened 2020-12-30
KH2 Text Editor improvements

Xeeynamo:
Brings the following changes:

* Say goodbye to `{:newstring}` as you now just need to press Enter to insert a new line.
* If opening a `sys.bar`, autodetects if European or Japanese encoding needs to be used.
* When opening a message file, autodetects `fontimage.bar` and `fontinfo.bar` rather than loading then manually.
* Updates documentation according to the latest changes.


Rikux3 (approved):
Couldn't find any problems.

PR #345, opened 2020-12-31
Sumn update

osdanova:


docs/kh2/file/type/00battle.md
```diff
@@ -481,7 +481,11 @@ Data on summons.
 
 | Offset | Type  | Description
 |--------|-------|--------------
-| 00     | 2B[7] | ???
+| 0     | 2B | ???
```
Xeeynamo, line 484: For the sake of consistency, could you change `2B` as `short`/`ushort` and `50B` as `byte[50]`?

Xeeynamo (changes requested):

docs/kh2/file/type/00battle.md
osdanova: Done

Xeeynamo (approved):

Issue #348, opened 2021-01-03
IdxImg KH2US support


DanielSinth:
Hi guys, first of all thank you very much for the amazing work you have done so far.
May I ask you to add support for KH2US.iso for the IdxImg tool?

OpenKh.Command.IdxImg inject -idx 3825664 KH2US.iso <InputFile> <FilePath>
FATAL ERROR: Specified argument was out of the range of valid values. (Parameter 'count')

ISO block for KH2.IDX. By default is 1840 for KH2FM
**for KH2US (0x3A6000, dec 3825664)**
ISO block for KH2.IMG. By default is 671693 for KH2FM
**for KH2US (0x459000, dec 4558848)**

Topaz-Reality:
I will look into this but it seems that the values you specified are too big for KH2US, perhaps?

Rikux3:
IdxImg uses specific values for 2FM currently, which is why any other version isn't supported yet.

DanielSinth:
@Topaz-Reality
> I will look into this but it seems that the values you specified are too big for KH2US, perhaps?

![IDX](https://user-images.githubusercontent.com/38701288/103487208-c5379a00-4e14-11eb-8dfc-c4d048a0852a.jpg)

Topaz-Reality:
@DanielSinth 
Yea, it is asking you for **blocks**, so you need to divide the value you input by 2048.
So, 3825664 / 2048 = **1868** & 4558848 / 2048 = **2226**

I will implement a dynamic file seeking method to this so manual labor is not necessary. But in the meantime, please use the given values above.

DanielSinth:
Thanks a lot!
Now it works without any problems.
Should I close the issue?

Topaz-Reality:
We should close the issue after and if #349 is tested and accepted. In my opinion, anyways.

Topaz-Reality:
Solved as of #349. Thanks for the issue!

PR #349, opened 2021-01-03
Somewhat hacky but effective way to implement IDX/IMG LBA Seek.

Topaz-Reality:
Does what the title says. Seeks out the LBA Offset of the IDX and IMG from the ISO.
This is hacky, but as far as I have seen, it works with everything tested.

Addresses #348.

Tested ISOs:
-> KH2US
-> KH2FM
-> KH2CCZ

Of course, more testing is necessary. But this will do until we have a proper ISO library.

### DIFF
--- a/OpenKh.Command.IdxImg/Program.cs
+++ b/OpenKh.Command.IdxImg/Program.cs
@@ -205,19 +205,6 @@ namespace OpenKh.Command.IdxImg
             [Option(CommandOptionType.SingleValue, Description = "ISO block for KH2.IMG. Determined automatically by default", ShortName = "img", LongName = "img-offset")]
             public int ImgIsoBlock { get; set; } = -1;
 
-            bool CompareArrays(byte[] input1, byte[] input2)
-            {
-                bool _return = true;
-
-                for (int i = 0; i < input1.Length; i++)
-                {
-                    if (input1[i] != input2[i])
-                        _return = false;
-                }
-
-                return _return;
-            }
-
             protected int OnExecute(CommandLineApplication app)
             {
                 const long EsitmatedMaximumImgFileSize = 4L * 1024 * 1024 * 1024; // 4GB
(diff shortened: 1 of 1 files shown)


OpenKh.Command.IdxImg/Program.cs
```diff
@@ -213,6 +226,49 @@ protected int OnExecute(CommandLineApplication app)
...
+                    isoStream.Position = (uint)(0x105 * 0x800);
+                    byte[] haystack = isoStream.ReadBytes(0x500);
+
+                    using (BinaryReader reader = new BinaryReader(new MemoryStream(haystack)))
+                    {
+                        for (int i = 0; i < 0x500; i++)
+                        {
+                            reader.BaseStream.Position = i;
+
+                            byte[] hayRead = reader.ReadBytes(0x0B);
```
Xeeynamo, line 245: Suggestion here to replace `0x0B` with a constant (eg. `NeedleLength`) and use it when declaring those arrays above (eg. `imgNeedle = new byte[NeedleLength]` so the reason why it is required to read that amount of byte is more explicit.

OpenKh.Command.IdxImg/Program.cs
```diff
@@ -213,6 +226,49 @@ protected int OnExecute(CommandLineApplication app)
...
+
+                if (IdxIsoBlock == -1 || ImgIsoBlock == -1)
+                {
+                    byte[] imgNeedle = new byte[] { 0x01, 0x09, 0x4B, 0x48, 0x32, 0x2E, 0x49, 0x4D, 0x47, 0x3B, 0x31 };
+                    byte[] idxNeedle = new byte[] { 0x01, 0x09, 0x4B, 0x48, 0x32, 0x2E, 0x49, 0x44, 0x58, 0x3B, 0x31 };
+
+                    isoStream.Position = (uint)(0x105 * 0x800);
+                    byte[] haystack = isoStream.ReadBytes(0x500);
+
+                    using (BinaryReader reader = new BinaryReader(new MemoryStream(haystack)))
```
Xeeynamo, line 239: I believe it is not required to read a `0x500` block and using it in the `BinaryReader`. Can you do `new BinaryReader(isoStream)` instead? If you are concerned about performance, you can wrap `isoStream` with a `BufferedStream`.

OpenKh.Command.IdxImg/Program.cs
```diff
@@ -199,11 +199,24 @@ private class InjectCommand
...
-            public long IdxIsoBlock { get; set; } = 1840;
+            [Option(CommandOptionType.SingleValue, Description = "ISO block for KH2.IDX. Determined automatically by default.", ShortName = "idx", LongName = "idx-offset")]
+            public int IdxIsoBlock { get; set; } = -1;
 
-            [Option(CommandOptionType.SingleValue, Description = "ISO block for KH2.IMG. By default is 671693 for KH2FM", ShortName = "img", LongName = "img-offset")]
-            public long ImgIsoBlock { get; set; } = 671693;
+            [Option(CommandOptionType.SingleValue, Description = "ISO block for KH2.IMG. Determined automatically by default", ShortName = "img", LongName = "img-offset")]
+            public int ImgIsoBlock { get; set; } = -1;
+
+            bool CompareArrays(byte[] input1, byte[] input2)
```
Xeeynamo, line 208: No need to create a dedicated function. `input1.SequenceEqual(input2)` will be good enough and it's fast when comparing two arrays of the same length!

OpenKh.Command.IdxImg/Program.cs
```diff
@@ -213,6 +226,49 @@ protected int OnExecute(CommandLineApplication app)
...
+
+                            if (imgCmp || idxCmp)
+                            {
+                                reader.BaseStream.Position -= 0x24;
+
+                                byte[] blockStack = reader.ReadBytes(4);
+                                byte[] blockCorrect = new byte[] { blockStack[3], blockStack[2], blockStack[1], blockStack[0] };
+
+                                if (idxCmp && IdxIsoBlock == -1)
+                                    IdxIsoBlock = BitConverter.ToInt32(blockCorrect);
```
Xeeynamo, line 258: I never considered to use `BitConverter.ToInt32`. Thanks for sharing this tip :)

OpenKh.Command.IdxImg/Program.cs
```diff
@@ -213,6 +226,49 @@ protected int OnExecute(CommandLineApplication app)
 
                 using var isoStream = File.Open(InputIso, FileMode.Open, FileAccess.ReadWrite);
 
+                // Begin being hacky.
```
Xeeynamo, line 229: No need to specify this is an hack as long as it works ^^

Xeeynamo (changes requested):
Small tip: use `var` instead of the explicit type when possible

Xeeynamo:
Another suggestion: if you still plan to use that buffer of `0x500` you can actually get rid of `BinaryReader` and the `ReadBytes` all together. .NET Core 3.1 comes with `Span<T>`, so you can slice a portion of an array with no performance impact. For instance, those two expressions are functionally equivalent, but the second one is more performant:

```csharp
var reader = new BinaryReader(new MemoryStream(buffer));
reader.BaseStream.Position = 5;
var data = reader.ReadBytes(11)
```
with Span<T>:
```csharp
var data = new Span(buffer, 5, 11);
```

Topaz-Reality:
Changes made.

Xeeynamo (approved):

Issue #342, opened 2020-12-29
MapStudio Crash


1234567890num:
MapStudio crashes when trying to open b_00 spawn point in eh06 (Nothing's Call)

Topaz-Reality:
Could you provide the error messages that pop up (if any) and some screenshots to help the devs understand the issue?
Thanks a ton for the issue.

1234567890num:
There's no error messages before the crash. In the scenario presented in the screenshot the progam will crash immediately upon clicking.

![crash](https://user-images.githubusercontent.com/53823593/103334268-4d2f4400-4aa3-11eb-9d0d-a6f7aee6e893.png)

Xeeynamo:
It seems to be caused by `obj\M_EX870.mdlx` as MapStudio (`ObjEntryController.cs:74`) always expect a texture. The 3D model is a white pyramid and it looks like to be unused in the final game too.

Vladabdf:
Fixed with #350.

PR #350, opened 2021-01-03
Fix MapStudio crashing when it loads a model with no textures.

Topaz-Reality:
A quick fix to #342. It's a fast fix, but it's an effective one.

Screenshot:
image


Xeeynamo (changes requested):
As catching an exception is an expensive operation and it would be better to explicitly fail for other uncaught bugs, I suggest to do this is instead:
```csharp
var textureEntry = mdlxEntries.FirstOrDefault(x => x.Type == Bar.EntryType.ModelTexture);
if (textureEntry != null)
    textures = ModelTexture.Read(textureEntry.stream);
```
The rest looks very fine. Thanks a lot for the improvement 👍

Topaz-Reality:
Roger. Allow me.

Topaz-Reality:
The requested change has been made.

Xeeynamo (approved):

PR #362, opened 2021-01-14
Birth by Sleep PMO support

Truthkey:
Adds PMO support to the engine:

* Can read and fully display any PMO file from Birth by Sleep.
* Can write PMO files given the right data.

_Screenshot of PMO model loaded on OpenKH.Game_
TerraOpenKH

Example of usage of the loader function below.
```
(Model, Textures) = BBSMeshLoader(graphics, "model/h03ex00.pmo", Model, Textures);
```

**Known Issues**
- Models that use textures that are not fully square have distorted textures.
- No bone support.


OpenKh.Bbs/Pmo.cs
```diff
@@ -0,0 +1,716 @@
...
+            Pmo pmo = new Pmo();
+
+            ReadHeader(ref stream, ref pmo);
+            ReadTextureSection(ref stream, ref pmo);
+
+            // Read all data
+            UInt16 VertCnt = 0xFFFF;
+            int meshSecCount = 0;
+
+            ReadMeshData(ref stream, ref pmo, ref VertCnt, ref meshSecCount, 0);
```
Xeeynamo, line 497: I do not think you need `ref` to `stream` and `pmo`. And as you are always initialising `VertCnt` I think you can even inline that number as parameter.

OpenKh.Bbs/Pmo.cs
```diff
@@ -0,0 +1,716 @@
...
+                    pmo.jointList[j] = BinaryMapping.ReadObject<JointData>(stream);
+                }
+            }
+
+            return pmo;
+        }
+
+        public static void Write(Stream stream, Pmo pmo)
+        {
+            BinaryWriter wr = new BinaryWriter(stream);
```
Xeeynamo, line 537: I noticed you are using `BinaryWriter` to write some `sbyte` and `float` values. There are some helpers that do that for you in `OpenKh.Common` and there is not really need to use `BinaryWriter`. Just call `stream.Write(yourFloat)` and it should do it for you.

OpenKh.Bbs/Pmo.cs
```diff
@@ -0,0 +1,716 @@
...
+                for (int joint = 0; joint < pmo.jointList.Length; joint++)
+                {
+                    BinaryMapping.WriteObject<JointData>(stream, pmo.jointList[joint]);
+                }
+            }
+        }
+
+        public static bool IsValid(Stream stream) =>
+            stream.Length >= 4 &&
+            new BinaryReader(stream.SetPosition(0)).ReadUInt32() == MagicCode;
```
Xeeynamo, line 714: Same here, you do not need `BinaryReader` as `stream.SetPosition(0).ReadUInt32()` will just work.

OpenKh.Engine.MonoGame/KingdomTexture.cs
```diff
@@ -40,6 +43,30 @@ public void Dispose()
         }
     }
 
+    public class Tim2KingdomTexture : IKingdomTexture
+    {
+        private static readonly Vector2 DefaultRegion = new Vector2(0, 1);
+
+        public Tim2KingdomTexture(byte[] texture, GraphicsDevice graphics)
+        {
+            List<Tm2> tm2 = Tm2.Read(new MemoryStream(texture)).ToList();
```
Xeeynamo, line 52: To squeeze a little amount of performance, as you already know that the returned value from `Tm2.Read` will always have a single element, I suggest you to do `Tm2.Read(new MemoryStream(texture)).First()`.

OpenKh.Tests/Bbs/PmoTests.cs
```diff
@@ -0,0 +1,138 @@
...
+using Xunit;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System;
+
+namespace OpenKh.Tests.Bbs
+{
+    public class PmoTests
+    {
+        private static readonly string FileName = "Bbs/res/p01ex00.pmo";
```
Xeeynamo, line 13: I have a change request about this, but I will comment it on Discord for safety reasons.

OpenKh.Tests/Bbs/PmoTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            var TestPmo = Pmo.Read(stream);
+            Assert.Equal("Root", TestPmo.jointList[0].JointName);
+            Assert.Equal(0x1, TestPmo.jointList[1].JointIndex);
+        });
+
+        [Fact]
+        public void CheckH03EX00()
+        {
+            var TestPmo = Pmo.Read(File.OpenRead("../../../Bbs/res/h03ex00.pmo"));
```
Xeeynamo, line 108: Files inside `Bbs/res/` should always be copied. So you do not need to return back to the directory with `../`. This rule is defined in `OpenKh.Tests.csproj` file, line 36.

Xeeynamo (changes requested):
Very exciting times for OpenKH. Your contribution here is a massive step forward to show the capabilities of the OpenKH Game Engine. Before merging, I request some small changes here and there. I can not wait to see what we are going to see in the future thanks to this.

OpenKh.Bbs/Pmo.cs
Truthkey: I can definitely do without ```VertCnt```, but ```stream``` and ```pmo``` need to be ```ref``` or when you read the second mesh (that uses the same method) you won't be modifying the same object.

OpenKh.Bbs/Pmo.cs
Truthkey: ```stream.Write(yourFloat)``` doesn't seem to work, which is why I went for BinaryWriter. I have to do ```OpenKh.Common.StreamExtensions.Write(yourFloat)``` for it to work. Or without ```OpenKh.Common``` in my case, since I'm using ```using OpenKh.Common```.

OpenKh.Engine.MonoGame/KingdomTexture.cs
Truthkey: Made it even shorter, like this:
```
public Tim2KingdomTexture(byte[] texture, GraphicsDevice graphics)
{
   Texture2D = Tm2.Read(new MemoryStream(texture)).First().CreateTexture(graphics);
}
```

Xeeynamo (approved):

PR #365, opened 2021-01-14
Automatically merge PR from Dependabot

Xeeynamo:
As they are pretty chatty and annoying, I suggest to auto-merge them. Example of them are #363 and #364 .


Truthkey (approved):
This sounds like a logical move since they're mostly updates. Everything seems fine here.

PR #380, opened 2021-01-23
Add object collision parser and documentation

Xeeynamo:
Me, Luseu and Delta spent half a day to figure out how they work. This contains all our research so far.


Topaz-Reality (commented):
They seem good to me, nice work!
However, since they are incomplete, and thus in progress, I do not want to approve any changes by myself.

Xeeynamo:
The only incomplete information is the collision type in the docs. Unfortunately after hours of research we did not understand those missing values. They can be added in the future if someone will try to look into them again, but for now the collision library can be used to give collisions to the engine 👍 which was why the PR was made in the first place

Rikux3 (approved):
In that case i'll approve those very welcoming additions

PR #381, opened 2021-01-23
Enums for KH2 and BBS

osdanova:
Added some enums from Delta's BBS decompilation that may be useful. Changed a couple of names in KH2's docs to adapt to the official names.


Topaz-Reality:
For one reason or another, the tests have failed.
Can you rebase this PR just in case?

Xeeynamo (changes requested):
I think it is better to move the `ATK_ATTR` in `00battle.md` rather than creating a dedicated page for it. Also I suggest to not use those names there as they belong to a leaked build... Let's avoid to copy & paste names from there when we can.

PR #385, opened 2021-01-23
AreaData further fields research

Xeeynamo:
Updates map studio and the documentation with the new discovered fields.


Truthkey (approved):
Really awesome changes. I gave the code a look and everything looks correct.

PR #389, opened 2021-01-25
Update MSN.MD

shinerkz:
Updated Bitflag Descriptions.


shinerkz:
> I've actually found that too and wanted to implement that with #321 .

Oh neat. Seems like yours has some different stuff too. But what are you still working on in that file?

Anyway we can let this go through for now and have it overwritten by #321 later when you're done?

Rikux3 (approved):

PR #392, opened 2021-01-25
creates an image folder for unused models

Tomk1004:
The folder is empty since I'm just making a base for the images, also, I don't think adding UI and textures to the official GitHub is a good idea due to potential legal issues, so I won't be adding those for the time being, though I could be wrong, so please correct me if I am


Topaz-Reality:
Can I request that you put this PR together alongside with your table for unused content? You should be able to create and use folders with markdown in the same PR. One PR for one folder addition seems excessive.

Topaz-Reality (changes requested):
Can I request that you put this PR together alongside with your table for unused content? You should be able to create and use folders with markdown in the same PR. One PR for one folder addition seems excessive.

Truthkey (changes requested):
I'd request changing the folder name so it doesn't have any spaces. Either just calling it `unused` or adding an underscore.

Tomk1004:
I didn't include the table with the unused images because I was planning on referencing images that were going to be put into this folder

Tomk1004:
> 
> 
> I'd request changing the folder name so it doesn't have any spaces. Either just calling it `unused` or adding an underscore.

will do

Vladabdf:
> Can I request that you put this PR together alongside with your table for unused content? You should be able to create and use folders with markdown in the same PR. One PR for one folder addition seems excessive.

+1, just do it all in one PR.

Screenshots of assets, such as unused models are fine, but yes if it entails uploading official textures, it would be best to avoid doing that.

Xeeynamo:
If you would like to use the placeholder image, without merging this to master, you could use the following url:
https://raw.githubusercontent.com/Xeeynamo/OpenKh/207bd4260099c9bf20074622c92b16b32b6f8982/docs/kh2/images/unused/placeholder.png

Topaz-Reality:
Closing this PR. PRing is unnecessary for just a single folder.

PR #394, opened 2021-01-25
Add script to automatically format code before merging PRs

Xeeynamo:
I was planning this PR since months, but I never pushed it as I knew it would cause a major pain at rebasing and merging. But as lately more and more people are forking and making changes to the project, I can not really delay this any further.

This PR adds an Action that automatically creates commits that formats the code based on the specifications in the [.editorconfig](https://github.com/Xeeynamo/OpenKh/blob/master/.editorconfig). The commits that fixes the formatting will look like [this](https://github.com/Xeeynamo/KH3SaveEditor/pull/120).


Xeeynamo:
I am marking this PR as ready to be reviewed. There are no many reasons to hold this back any longer. 

I highly suggest to merge this only after #457 , #416 and #458 to avoid conflicts.

Vladabdf:
Closing, but the branch will remain in the event this is actually useful and can be cleaned and merged later on if necessary. (It seems pedantic to me.)

PR #401, opened 2021-01-26
Standardise OpenKH Log across the entire codebase

Xeeynamo:
Move `Log` from `OpenKh.Game` to `OpenKh.Common`, so every project can use it by just doing `Log.Error("MyError")` without any set-up. The log takes the name of the executable file, so if you log from the app `OpenKh.Tools.MapStudio` the log will be called `OpenKh.Tools.MapStudio.log`. You do not need to deal with any configuration, _just log™_

Another feature is that you can hook incoming log messages. Just do a `Log.OnLogDispatch += MyLogHandler` to get them. That way you can for example display log messages into an application.

### DIFF
--- a/OpenKh.Common/Log.cs
+++ b/OpenKh.Common/Log.cs
@@ -11,7 +11,7 @@ namespace OpenKh.Common
 {
     public static class Log
     {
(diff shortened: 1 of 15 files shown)


Rikux3 (approved):
A very welcome change!

PR #408, opened 2021-01-27
Remove monogame dependency and slightly improve performance

Xeeynamo:
To manage matrices and vectors, we currently have both `System.Numerics` and `Microsoft.Xna.Framework`. `System.Numerics` is faster and with the latest versions of .NET Core, the compiler can generate specialised instructions for certain processors to boost performance even more. But `Microsoft.Xna.Framework`, which is part of MonoGame, have basically the same functionality but slower.

This PR aims to remove some overhead caused by the MonoGame library, to standardise the math library following the .NET Core specifications and to remove the dependency of it here and there.

### DIFF
--- a/OpenKh.Engine.MonoGame/Camera.cs
+++ b/OpenKh.Engine.MonoGame/Camera.cs
@@ -133,7 +133,7 @@ namespace OpenKh.Engine.MonoGame
             }
         }
 
-        public System.Numerics.Matrix4x4 Projection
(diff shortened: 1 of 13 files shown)


Topaz-Reality:
I do have experience with MonoGame. Will be reviewing in an hour or so.

Topaz-Reality (approved):
All seems to be fine.

PR #410, opened 2021-01-29
Debug skeleton drawing

thamstras:


Xeeynamo (approved):
Nice and clean. This will be very helpful with the incoming animation support from KH1 and BBS or the animation tweaks of KH2. Thanks for your first code contribution here in the OpenKH codebase ✌

PR #414, opened 2021-02-02
Documentation Guidelines

Truthkey:
Draft to discuss the documentation guidelines to follow when contributing new documentation to the project.

### DIFF
--- a/DOCS_GUIDELINE.md
+++ b/DOCS_GUIDELINE.md
@@ -1,0 +1,72 @@
+# Documentation guideline
+
+## General guidelines
+
+* Always use `hexadecimal` when defining offsets. They can be defined as follows. 
+  * With `0x` before the value: `0x1234`
+  * With `h` after the value: `1234h`
+
+* Although `hexadecimal` is preferred, `decimal` can be used for the rest of values.
+
+---
+
+## Representing single variables
+
+Here's a list of the recommended basic variable types. Size is in *bits*.
+
+| Size | Signed | Unsigned
+|------|--------|----------
+| 8  |  int8  |  uint8  
+| 16 |  int16 |  uint16
+| 32 |  int32 |  uint32
+| 32 |  float |  -
+| 64 |  int64 |  uint64
+| 64 |  double |  -
+
+If the variable contains actual text, `char` is preferred to `uint8` or `int8`.
+
+Arrays must be represented with the **[]** symbol. For example `int8[4]` means 4 int8 variables.
+The number of items in an array must be defined in `decimal` unless it is much more convenient to read in `hexadecimal`.
+
+Here are a few other special types you should take in account:
+
+| Size | Name | 
+|------|--------|
+| 64 |  Vector2f  |   
+| 96 |  Vector3f |  
+| 128 |  Vector4f |  
+| 512 |  Matrix4x4 |  
+
+---
+
+## Representing bitfields
+
+Bitfields always have to have their relative position and field size expressed in **decimal**.
+
+| Position | Size | Description
+|------|--------|----------
+|   0  |  1  | Field 1
+|   1  |  7  | Field 2
+|   8  |  8  | Field 3
+|   16  |  8  | Field 4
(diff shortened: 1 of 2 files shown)


Topaz-Reality:
This needs to be discussed before any more documentation is made. Henceforth:

Xeeynamo:
Honestly I like everything about those guidelines and I would not change a single thing. I'd say to fix the minor formatting issues (I can see a `---` at the end of every category), link it on the README or contributor guideline thing and then I would happily approve this!

DOCS_GUIDELINE.md
```diff
@@ -0,0 +1,72 @@
...
+
+---
+
+## Representing single variables
+
+Here's a list of the recommended basic variable types. Size is in *bits*.
+
+| Size | Signed | Unsigned
+|------|--------|----------
+| 8  |  int8  |  uint8
```
Topaz-Reality, line 19: I would prefer 'byte', 'short' and 'int' over 'int8', 'int16' and 'int32'

DOCS_GUIDELINE.md
```diff
@@ -0,0 +1,72 @@
...
+| 64 |  Vector2f  |   
+| 96 |  Vector3f |  
+| 128 |  Vector4f |  
+| 512 |  Matrix4x4 |  
+
+---
+
+## Representing bitfields
+
+Bitfields always have to have their relative position and field size expressed in **decimal**.
```
Topaz-Reality, line 44: A reason for this to be decimal? Positions are better presented by hexadecimal, wouldn't you think?

Topaz-Reality:
I made some comments to the guidelines. I would love to have people's opinions on my thoughts.

DOCS_GUIDELINE.md
Truthkey: The reason why I chose this is because it clearly states the amount of bytes it uses and the type of number it stores: `int8` is an integer 8 bits. `byte` can be confusing as to whether it includes signed values or not where `int8` has 0 ambiguity.

DOCS_GUIDELINE.md
Truthkey: Bitfields will rarely go over 32 bits in total and 99% of programming definitions of bitfields use decimal to specify the length.
For example:

```
struct bField
{
   int a:1;
   int b:8;
   int c:7;
}
```

So, I think using decimal makes more sense.

DOCS_GUIDELINE.md
Topaz-Reality: I mean, byte is one byte. short is 2 bytes. int is 4 bytes. This must be common knowledge.
Also the ambiguity talk is invalid because if there is no 'u' in front, it is signed.
But we should also get input from the others about this.

DOCS_GUIDELINE.md
Topaz-Reality: Oh I see. Very well then.

DOCS_GUIDELINE.md
Delta-473: Depending on compiler and target platform the size of an int can change. Personally I prefer int8, int16 etc over just short, int, long as there's no confusion possible.

Vladabdf: +1 to Delta's notion.

Topaz-Reality: That settles that, I guess.

osdanova: While I think byte,short,int... is easier to read, I agree with Delta, IntX/UintX specifies the exact size of the field, which I think works best for documentation. As a side note, @Truthkey, the same way you have int/uint, you also have sbyte/byte.

Topaz-Reality:
After some of the conversations made, it all seems good to me.

Xeeynamo:
Could you link this document in `CONTRIBUTING.md`?

Xeeynamo (approved):

Vladabdf (approved):

PR #416, opened 2021-02-04
KH2: Update 00battle, 03system and objentry editors

Rikux3:
Massive update on the source code files, as the documentation is much more advanced now.

The docs are now newer than the source code, so the tools should be updated at some point.

### DIFF
--- a/OpenKh.Tools.Kh2BattleEditor/ViewModels/PlrpViewModel.cs
+++ b/OpenKh.Tools.Kh2BattleEditor/ViewModels/PlrpViewModel.cs
@@ -1,0 +1,104 @@
+using OpenKh.Kh2;
+using OpenKh.Kh2.Battle;
+using OpenKh.Tools.Common.Models;
+using OpenKh.Tools.Kh2BattleEditor.Extensions;
+using OpenKh.Tools.Kh2BattleEditor.Interfaces;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Windows;
+
+namespace OpenKh.Tools.Kh2BattleEditor.ViewModels
+{
+    public class PlrpViewModel : MyGenericListModel<PlrpViewModel.PlrpEntryViewModel>, IBattleGetChanges
+    {
+        public class PlrpEntryViewModel
+        {
+            public Plrp Plrp { get; }
+
+            public PlrpEntryViewModel(Plrp plrp)
+            {
+                Plrp = plrp;
+            }
+
+            public string Name => $"{Plrp.Id:X02}";
+
+            public ushort Id { get => Plrp.Id; set => Plrp.Id = value; }
+            public byte Character { get => Plrp.Character; set => Plrp.Character = value; }
+            public byte Hp { get => Plrp.Hp; set => Plrp.Hp = value; }
+            public byte Mp { get => Plrp.Mp; set => Plrp.Mp = value; }
+            public byte Ap { get => Plrp.Ap; set => Plrp.Ap = value; }
+            public byte Strength { get => Plrp.Strength; set => Plrp.Strength = value; }
+            public byte Magic { get => Plrp.Magic; set => Plrp.Magic = value; }
+            public byte Defense { get => Plrp.Defense; set => Plrp.Defense = value; }
+            public byte ArmorSlotMax { get => Plrp.ArmorSlotMax; set => Plrp.ArmorSlotMax = value; }
+            public byte AccessorySlotMax { get => Plrp.AccessorySlotMax; set => Plrp.AccessorySlotMax = value; }
+            public byte ItemSlotMax { get => Plrp.ItemSlotMax; set => Plrp.ItemSlotMax = value; }
+
+            //display the starting items in a datagrid or similar...
+        }
+
+        private const string entryName = "plrp";
+
+        private string _searchTerm;
+
+        public PlrpViewModel(IEnumerable<Bar.Entry> entries) :
+            this(Plrp.Read(entries.GetBattleStream(entryName)))
+        { }
+
+        public PlrpViewModel() :
+            this(new List<Plrp>())
+        { }
+
+        private PlrpViewModel(IEnumerable<Plrp> items) :
+            base(items.Select(Map))
+        {
+        }
+
+        public string EntryName => entryName;
+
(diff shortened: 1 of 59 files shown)


Rikux3:
This PR is getting too big, so i'm putting this up for review now. 00battle and 00objentry documents should also be compliant with our documentation guidelines now. The rest can be done later.
@osdanova @1234567890num if you wouldn't mind to look over this (and of course everyone else who wants to)

docs/kh2/file/type/jiminy.md
```diff
@@ -120,28 +130,40 @@ Magic Code `JMAN`
...
-### Structure 2
+| Offset | Type   | Description |
+|--------|--------|-------------|
+| 00     | ushort | Label
+| 02     | short  | X Position
+| 04     | short  | Y Position
+| 06     | byte   | Draw
+| 07     | byte   | Padding 
 
 Currently unknown. Read in z_un_002a18d0.
```
osdanova, line 163: I take it this line was for "structure 2", which is now known.

docs/kh2/file/type/00objentry.md
```diff
@@ -6,76 +6,111 @@ Contains a definition of every object and it's parameters.
...
+| 0x54   | uint8 | Spawn Limiter
+| 0x55   | uint8 | Page (unknown)
+| 0x56   | uint8 | [Shadow Size](#shadow-size)
+| 0x57   | uint8 | [Form](#form-fm)
+| 0x58   | uint16 | Spawn additional object 1 - [OBJ LIST](../../dictionary/obj.md)
+| 0x5A   | uint16 | Spawn additional object 2 - [OBJ LIST](../../dictionary/obj.md)
+| 0x5C   | uint16 | Spawn additional object 3 - [OBJ LIST](../../dictionary/obj.md)
+| 0x5E   | uint16 | Spawn additional object 4 - [OBJ LIST](../../dictionary/obj.md)
+
+*The first bit is somewhat related to the enemy state. Eg: Undead Pirates lose their immunity if they are changed from 2 to 0
```
osdanova, line 31: This line is no longer needed

OpenKh.Tools.Kh2BattleEditor/ViewModels/PlrpViewModel.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        //public string SearchTerm
+        //{
+        //    get => _searchTerm;
+        //    set
+        //    {
+        //        _searchTerm = value;
+        //        PerformFiltering();
+        //    }
+        //}
```
osdanova, line 71: Guess these commented out pieces of code are non functional yet. I don't think it should be merged unless you plan on implementing it in the near future.

osdanova (changes requested):
Tools work as intended. Some changes I suggest before merging:

Enmp: Update the fields in the view.
image

Bons: Remove "Unknown", which is padding.
image

Sumn: Nothing is shown in the list, which is confusing. Can we add something there? Like the command's Id. Also, the field "ID" doesn't exist.
image

Objentry: Since you addded flag checkboxes, the "Flag" field should be removed.
image

docs/kh2/file/type/00battle.md
```diff
@@ -650,17 +732,19 @@ Contains the level-up table for summons and drive forms.
...
 | Offset | Type  | Description
 |--------|-------|-------------
-| 00     | byte  | First digit is the Form id, second digit is the Form level (e.g. 0x13 is Valor Form Level 3)
-| 01     | byte  | Level of the movement ability in the form (High Jump, Quick Run etc.)
-| 02     | short | Ability obtained through level up - [ITEM/ABILITY LIST](../../dictionary/inventory.md)
-| 04     | int   | EXP needed for level up
+| 00     | uint8 : 0-3 | [Id](#forms)
+| 00     | uint8 : 4-7 | Level
+| 01     | uint8 : 0-3 | Ability Level
+| 01     | uint8 : 4-7 | Anti Rate
```
osdanova, line 745: Anti rate and Ability Level are inversed. (Anti = 0-3; Abi = 4-7)

1234567890num: What are Ability Levels?

osdanova: Growth Ability levels for each form.
Eg: Valor form's level 1 and 2 have a Lvl 1 High Jump.

Rikux3: grafik
Am i confusing something?

osdanova: I'll show you Valor data so you can see:
image

And you wrote it right for the tool (Note that the function writes from right to left, so "4,4" are the first 4 bits):
image
image
image
image

osdanova (approved):
Docs are alright and the updated tools work as intended. Looks good to me.

Rikux3:
I've updated 03system's item table according to the decompilation. I would like for someone to verify this because it was confusing me a bit. I've also fixed the swapped events.
@1234567890num could you elaborate what you mean with the memt table?

1234567890num:
From ARD documentation:

#### Party

Set how the party member needs to be structured. According to `dbg/member.bin`, those are the only allowed values as a parameter:

| Value | Name          | Description
|-------|---------------|-------------
| 0     | NO_FRIEND     | Allow only Sora
| 1     | DEFAULT       | Allow Donald and Goofy
| 2     | W_FRIEND      | Allow Donald, Goofy and world guest.
| 3     | W_FRIEND_IN   | Same as `W_FRIEND` but forces the world guest to be in at the beginning.
| 4     | W_FRIEND_FIX  | Same as `W_FRIEND_IN`, but you can not swap the world guest.
| 5     | W_FRIEND_ONLY | Allow only with Sora and the world guest.
| 6     | DONALD_ONLY   | Only allow Sora and Donald.

The MEMT party table is what actually defines what these values do. For example, changing entry 0 of the table into 0x12030200 will give you Sora, Goofy, and Guest every time the game calls NO_FRIEND.

Rikux3:
I would really like to merge this now.

Delta-473:
Didn't had time to test it yet, but LGTM also needs rebase for the merge conflict.

Rikux3:
I have resolved the issues on 00battle.

Delta-473:
Can you look into why the tests broke, you might need to update those if you changed some names.

Rikux3:
Closing this in favor of #489

PR #420, opened 2021-02-11
Fix build script when running twice

Xeeynamo:
Basically redirects the `OpenKh.Winshell.Output` into `./bin` folder so it does not obstruct the `build.ps1` script. I modified the documentation of WinShell acoordingly other than indexing it on the docs.


Topaz-Reality (approved):
Nice PR, I shall blaze to it! I would have reviewed this at 4:20 but I could not wait that long.
Approved, thanks for fixing my stupid mistake with this blazing hot PR.
Keep on Smokin'

Delta-473:
PR looks good for me. At least I get to review PR420 ;)

PR #418, opened 2021-02-07
Very basic gameplay implementation

Xeeynamo:
This pull request aims to give to OpenKH Game Engine some basic gameplay mechanics:

https://user-images.githubusercontent.com/6128729/107161251-57423d80-6993-11eb-8692-90c6b3114c8c.mp4

While the character movement is entirely taken by reading values from PCSX2 in real time, the camera system is entirely built inside the engine. I will split the requirements of this PR into required and welcomed changes. While I am not sure to satisfy all the welcome changes before marking this as ready, I will do my best to complete the required changes.

Required changes:
* [X] Detect player entity
* [X] Camera should attach to player entity
* [X] Camera should move around the player
* [X] Accurately set FOV and aspect ratio exactly as the game does
* [x] Slowly adjust the horizontal rotation when the player moves
* [x] Slowly adjust the vertical rotation when the player moves
* [X] Interpolate rotation movement
* [x] Interpolate camera movement and FOV change
* [X] Implements AreaInfo from 03system
* [X] Adjust the camera based on the in-door or out-door value

Welcomed changes:
* [X] Read object collision data
* [X] Read map field collision
* [X] Read map camera collision
* [X] Read map color collision
* [ ] Implement camera collision
* [ ] Implement character field collision
* [ ] Implement diffuse color for objects
* [ ] Implement character color collision
* [x] Implement character movement
* [x] Set correct animation when character moves

PR #423, opened 2021-02-14
New input system

Xeeynamo:
Very inspired to the previous `InputManager`, this one is a re-modernisation of the previous one. You are still allowed to give a list of input devices, but all the logic to calculate the buttons repetition, released keys and triggered keys will be done after aggregating all the inputs rather than individually. Also I **FINALLY** created the interface `IInput`, so we can abstract the input implementation away from MonoGame. 

A part from the internal and technical changes, this is a list of features it brings to the table:
* Full analog support
* Pressed, Released, Triggered and Repeated keys
* Repeated keys are 1:1 with KH2
* Keyboard bindings can be, for now, configurable by editing the code

This will most likely break #418 as it relies on the old input system to move camera and player. Based on which one will be merged first, the other one needs to be rebased.


Topaz-Reality:
Since you have mentioned it:
May I ask you to, if it would be appropriate, to include this PR inside of #418? It would be better to have all things gameplay related (Yes, even the input system) in that mega PR to avoid issues. Just in case.

Xeeynamo:
> Since you have mentioned it:
> May I ask you to, if it would be appropriate, to include this PR inside of #418? It would be better to have all things gameplay related (Yes, even the input system) in that mega PR to avoid issues. Just in case.

I was thinking about that, but I prefer to have them reviewed and merged separately to have more continuous integration and detect potential bugs individually

Rikux3:
I haven't had a chance to test it yet, but from looking at the code I don't see anything wrong with it.

Xeeynamo:
I rebased the PR upon #418 as it was causing conflicts. I took the chance to add a smoother camera and character movement thanks to the analog values.

Truthkey (changes requested):
I would recommend changing the inputs to "brand" agnostic. For example, Cross would be FaceDown.
Same would go for the rest of buttons.

Xeeynamo:
> I would recommend changing the inputs to "brand" agnostic. For example, Cross would be FaceDown.
> Same would go for the rest of buttons.

Done 👍

Truthkey (approved):

Rikux3:
Plays smoothly and works flawless with my DS4.

PR #428, opened 2021-02-16
Add "Feature request" template

Vladabdf:
This adds a template similar to what Topaz created for bug reports, but for feature requests instead. Figured I might as well since I may be requesting some features in the near future.


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: "[FEATURE] - Requested feature"
+labels: 'feature'
```
Topaz-Reality, line 5: This should be 'enhancement', not 'feature'.

Topaz-Reality (changes requested):
Fix this one thing and it's in.

.github/ISSUE_TEMPLATE/feature_request.md
Vladabdf: Done.

Topaz-Reality (approved):

PR #434, opened 2021-02-18
Fix crash when deleting all layers in the Layout Editor

### LINKED ISSUE #396: Layout Editor crashes when deleting an animation's last layer
As title says. This used to not happen before the recent update. 
However, it is also impossible to add a new animation layer if the layer count = 0. Both of these should be fixed ASAP, both in the name of optimization and to be bug-free.

Xeeynamo:
Fixes #396


Topaz-Reality (commented):
Seems good. Approved!

Vladabdf (approved):
Perfect, works as intended now. I can remove all layers and the program doesn't crash, and I can increase the layer count from 0. Thanks papa Xeey. 👍

PR #440, opened 2021-02-19
Changes the bar file type 2 to OtherData

gjgfuj:
Changes the bar file type 2 to OtherData which provides more clarity since these files don't have a known type although they may generally be lists.

### DIFF
--- a/OpenKh.Command.Bar/Helpers.cs
+++ b/OpenKh.Command.Bar/Helpers.cs
@@ -12,7 +12,7 @@ namespace OpenKh.Command.Bar
             {
                 [EntryType.Dummy] = "dummy",
                 [EntryType.Binary] = "bin",
-                [EntryType.List] = "list",
+                [EntryType.OtherData] = "list",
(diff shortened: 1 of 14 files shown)


Xeeynamo:
I was thinking about it and `Binary` might be more appropriate than `OtherData`

Topaz-Reality:
> I was thinking about it and `Binary` might be more appropriate than `OtherData`

Binary does not fit this description. I think 'list' already suits it as well as it can.

gjgfuj:
type 0x01 is for BAR files. so the existing Binary type is a misnomer too. Binary is a solid name too, and that'd be perfectly valid if that's what you'd wish to change it to.

Topaz-Reality:
Upon further inspection, the list of names used in the BAR editor need to be refined and renewed.
I propose for this PR's closure, and creation of a project/issue/request where the list is renewed in a way that all collaborators can agree upon and that it would be accurate to the game.
I won't be closing the PR myself, however.

PR #443, opened 2021-02-22
PMO list and KH2 enemy list update

osdanova:
Added a Dictionary section to BBS's docs. Worlds file moved to dictionaries and PMO dictionary added.
Updated the enemy list for KH2's battle editor.


Truthkey (changes requested):
In `models.md` make sure to separate the models by category. Add subsections with ### to separate B/G/F and the like. Also, if possible have links with [id] (#Category-you-want) without the space in the Category section.

osdanova:
Didn't divide it at first so we had a continuous list that can be copied and reordered, but it's easy enough to just remove the sections if needed when doing so. Done.

docs/bbs/index.md
```diff
@@ -16,3 +15,7 @@
 * Other files
     * [font.arc, fonten.arc](font.md)
 
+## Dictionaries
+
+* [Worlds](dictionary/worlds.md)
```
Delta-473, line 20: Is moving worlds.md to the dictionary dir coming in an other commit?

osdanova: I actually forgot... Thank you.

Truthkey:
You are still missing a lot of models in this list. For example, none of the **P** from **BD** world are included.
For example: P01BD00, P51BD00...

Also, I recommend you to include the 2.5 exclusive P93EX00 which is Armored Terra with a cape.

osdanova:
> You are still missing a lot of models in this list. For example, none of the **P** from **BD** world are included.
> For example: P01BD00, P51BD00...
> 
> Also, I recommend you to include the 2.5 exclusive P93EX00 which is Armored Terra with a cape and the JP exclusive W14VE00 which is Light Seeker.

I've added Terra, thank you! As for the rest:
About light seeker, are you sure that's the id? It's the same filename as Frolic Flame. (2FM)
About P01BD00, P51BD00... those are arc files and not pmo files, thus I have not listed them.

Could you explain these 2 points to me in the Discord Server? Thank you.

osdanova:
Light Seeker added. Change "Models" to "PMO files" because arc files also contain models and are not listed.

Truthkey (approved):

PR #446, opened 2021-02-24
Optimize I/O performance

Xeeynamo:
Before to read some data (eg. `stream.ReadInt32()`) internally a `BinaryReader` object had to be created, which added some overhead.
With this PR, there is a direct memory mapping thanks to some low-level features from C#. I also introduced a `stream.ReadMatrix4x4` and a `stream.Write(Matrix4x4)` which also boost significantly the performance.

Unit tests on my PC used to become green in 25.8 seconds. With this PR they just take 21.3 seconds.


Truthkey (approved):
Your changes seem correct and seem to improve performance. Approved!

PR #449, opened 2021-02-26
Fix inverted character controls

Xeeynamo:
Fix inverted controls in both OpenKh.Game and OpenKh.Tools.MapStudio


Truthkey (approved):
I just tested it and it works correctly.

PR #452, opened 2021-03-02
Add "listreplace" method for patching

tommadness:
Adds the ability to modify certain KH2 lists based on a YAML file. The editable lists are:
- 03system.bin
- - Trsr
- - Item
- 00battle.bin
- - Fmlv
- - Lvup
- - Bons

Adds the ability to deobfuscate encoded YAML files using Base64 Encoding. This is intended to be used with KH2 Randomizer, so that the source files aren't human readable to prevent cheating.

### DIFF
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -59,6 +59,9 @@ namespace OpenKh.Patcher
 
         public void Patch(string originalAssets, string outputDir, Metadata metadata, string modBasePath)
         {
+            Directory.Delete(outputDir, true);
+            Directory.CreateDirectory(outputDir);
+
             var context = new Context(metadata, originalAssets, modBasePath, outputDir);
             try
             {
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -120,6 +120,9 @@ namespace OpenKh.Patcher
                 case "kh2msg":
                     PatchKh2Msg(context, assetFile.Source, stream);
                     break;
+                case "areadatascript":
+                    PatchAreaDataScript(context, assetFile.Source, stream);
+                    break;
             }
 
             stream.SetLength(stream.Position);
@@ -256,6 +259,19 @@ namespace OpenKh.Patcher
             }
 
             Msg.WriteOptimized(stream.SetPosition(0), msgs);
+        }
+
+        private static void PatchAreaDataScript(Context context, List<AssetFile> sources, Stream stream)
+        {
+            var scripts = Kh2.Ard.AreaDataScript.Read(stream).ToDictionary(x => x.ProgramId, x => x);
+            foreach (var source in sources)
+            {
+                var programsInput = File.ReadAllText(context.GetSourceModAssetPath(source.Name));
+                foreach (var newScript in Kh2.Ard.AreaDataScript.Compile(programsInput))
+                    scripts[newScript.ProgramId] = newScript;
+            }
+
+            Kh2.Ard.AreaDataScript.Write(stream.SetPosition(0), scripts.Values);
         }
     }
 }
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -59,7 +59,8 @@ namespace OpenKh.Patcher
 
         public void Patch(string originalAssets, string outputDir, Metadata metadata, string modBasePath)
         {
-            Directory.Delete(outputDir, true);
+            if (Directory.Exists(outputDir))
+                Directory.Delete(outputDir, true);
             Directory.CreateDirectory(outputDir);
 
             var context = new Context(metadata, originalAssets, modBasePath, outputDir);
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -59,10 +59,6 @@ namespace OpenKh.Patcher
 
         public void Patch(string originalAssets, string outputDir, Metadata metadata, string modBasePath)
         {
-            if (Directory.Exists(outputDir))
-                Directory.Delete(outputDir, true);
-            Directory.CreateDirectory(outputDir);
-
             var context = new Context(metadata, originalAssets, modBasePath, outputDir);
             try
             {
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -74,6 +74,9 @@ namespace OpenKh.Patcher
 
                     foreach (var name in names)
                     {
+                        if (assetFile.Required && !File.Exists(context.GetOriginalAssetPath(name)))
+                            continue;
+
                         context.CopyOriginalFile(name);
                         var dstFile = context.GetDestinationPath(name);
 
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -71,11 +71,19 @@ namespace OpenKh.Patcher
 
                 metadata.Assets.AsParallel().ForAll(assetFile =>
                 {
-                    context.CopyOriginalFile(assetFile.Name);
-                    var dstFile = context.GetDestinationPath(assetFile.Name);
+                    var names = new List<string>();
+                    names.Add(assetFile.Name);
+                    if (assetFile.Multi != null)
+                        names.AddRange(assetFile.Multi.Select(x => x.Name).Where(x => !string.IsNullOrEmpty(x)));
 
-                    using var stream = File.Open(dstFile, FileMode.OpenOrCreate, FileAccess.ReadWrite);
-                    PatchFile(context, assetFile, stream);
+                    foreach (var name in names)
+                    {
+                        context.CopyOriginalFile(name);
+                        var dstFile = context.GetDestinationPath(name);
+
+                        using var stream = File.Open(dstFile, FileMode.OpenOrCreate, FileAccess.ReadWrite);
+                        PatchFile(context, assetFile, stream);
+                    }
                 });
             }
             catch (Exception ex)
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -88,6 +88,7 @@ namespace OpenKh.Patcher
             }
             catch (Exception ex)
             {
+                Log.Err($"Patcher failed: {ex.Message}");
                 throw new PatcherException(metadata, ex);
             }
         }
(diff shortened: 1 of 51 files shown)


tommadness:
Test mod link: https://discord.com/channels/409140906625728532/810393223905411133/817009783706091560

- Changes Kingdom Key to a 7 / 7 with Second Chance #Item
- Changes Garden of Assemblage Map Chest to contain Valor Form #Trsr
- Changes experience for level 2 Valor to 1, reward to Fire Element #Fmlv
- Changes experience for level 2 Sora to 1, Strength and Magic to 10 #Lvup
- Changes Hollow Bastion Bailey "Get Bonus" to Wisdom Form #Bons

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,79 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+        private static void PatchListReplace(Context context, List<AssetFile> sources, Stream stream)
+        {
+            foreach (var source in sources)
+            {
+                var deserializer = new DeserializerBuilder().IgnoreUnmatchedProperties().Build();
+                Dictionary<string, byte> characterMap = new Dictionary<string, byte>() {
+                            { "Sora", 1 }, { "Donald", 2 }, { "Goofy", 3 },  { "Mickey", 4 },  { "Auron", 5 }, { "PingMulan",6 }, { "Aladdin", 7 },  { "Sparrow", 8 }, { "Beast", 9 },  { "Jack", 10 },  { "Simba", 11 }, { "Tron", 12 }, { "Riku", 13 }, { "Roxas", 14}
+                        };
+                    string sourceText = File.ReadAllText(context.GetSourceModAssetPath(source.Name));
+                    if (source.Name.Contains("enc"))
```
Xeeynamo, line 297: I know that encoding the file as base64 is to obfuscate files to not cheat on randomizers, but I prefer to not couple this functionality with the patcher functionality. So I'd say to remove the next 5 lines of code.
I was thinking on a way to randomize values on patching time rather than modfying the mod files. But it would be out of scope of this PR anyway.

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,79 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
 
             Kh2.Ard.AreaDataScript.Write(stream.SetPosition(0), scripts.Values);
         }
+
+        private static void PatchListReplace(Context context, List<AssetFile> sources, Stream stream)
+        {
+            foreach (var source in sources)
+            {
+                var deserializer = new DeserializerBuilder().IgnoreUnmatchedProperties().Build();
+                Dictionary<string, byte> characterMap = new Dictionary<string, byte>() {
```
Xeeynamo, line 293: This will be very inefficient as you are creating a dictionary at every loop. I recommend you to move this outside the method and prepend it with a `private static readonly`. Same for your deserializer.

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,79 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+                    {
+                        byte[] data = System.Convert.FromBase64String(sourceText);
+                        sourceText = System.Text.ASCIIEncoding.ASCII.GetString(data);
+                    }
+                    switch (source.Type)
+                    {
+                        case "trsr":
+                            var TrsrList = Kh2.SystemData.Trsr.Read(stream).ToDictionary(x => x.Id, x => x);
+                            var moddedTrsr = deserializer.Deserialize<Dictionary<int, Kh2.SystemData.Trsr>>(sourceText);
+                            foreach (KeyValuePair<int, Kh2.SystemData.Trsr> treasure in moddedTrsr)
```
Xeeynamo, line 307: A suggestion in C# is to use `var` whenever is possible. You can replace `KeyValuePair<int, Kh2.SystemData.Trsr>` with `var`, making it more readable. I am commenting it on this precise line of code, but I suggest to change to `var` on all the code changes you made.

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,79 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+                    string sourceText = File.ReadAllText(context.GetSourceModAssetPath(source.Name));
+                    if (source.Name.Contains("enc"))
+                    {
+                        byte[] data = System.Convert.FromBase64String(sourceText);
+                        sourceText = System.Text.ASCIIEncoding.ASCII.GetString(data);
+                    }
+                    switch (source.Type)
+                    {
+                        case "trsr":
+                            var TrsrList = Kh2.SystemData.Trsr.Read(stream).ToDictionary(x => x.Id, x => x);
```
Xeeynamo, line 305: Please use camel-case. Your `TrsrList` would become `trsrList`. This is to keep consistency with the rest of the code base.

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,79 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+                            {
+                                TrsrList[treasure.Value.Id].ItemId = treasure.Value.ItemId;
+                            }
+                                Kh2.SystemData.Trsr.Write(stream.SetPosition(0), TrsrList.Values);
+                            break;
+
+                        case "item":
+                            var ItemList = Kh2.SystemData.Item.Read(stream);
+                            var moddedItem = deserializer.Deserialize<Kh2.SystemData.Item>(sourceText);
+                            ItemList = moddedItem;
```
Xeeynamo, line 317: You're reading `itemList` from the source, although rather than patching it you are replacing its content. Is this intended?

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,79 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+                            ItemList.Write(stream.SetPosition(0));
+                            break;
+
+                        case "fmlv":
+                            var FormList = Kh2.Battle.Fmlv.Read(stream).ToDictionary(x => String.Concat(x.FormFm, x.FormLevel), x => x);
```
Xeeynamo, line 322: I would suggest to create a Dictionary where the key is the `FormFm` and the value is a list of `Kh2.Battle.Fmlv.Level`, where every index of the list represents a level. This would help to potentially add new levels to individual drive forms in the future. Although this might require a change to your YAML. In my mind it would become something like the following:
```yml
Wisdom:
  - FormLevel: 2
    Ability: 3
    Exp: 1234
```
where you can take the "Wisdom" name out of it to know in which index of the original file you have to patch the values. It would be also important to allow the same files to patch both Vanilla and the FinalMix versions of the game. Although I realise it might be complex.

There is also an update on how the `Fmlv` is parsed https://github.com/Xeeynamo/OpenKh/blob/f6471f0339f0a625ad69f0b4700efe50ae24f36a/OpenKh.Kh2/Battle/Fmlv.cs . I believe this is a complex problem to solve at the moment. Finally, to prevent to break the compatibility with future renaming to the parser, could you create a dedicated class that maps to the actual `Fmlv`?

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,79 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+                                FormList[form.Key].Exp = form.Value.Exp;
+                            }
+                            Kh2.Battle.Fmlv.Write(stream.SetPosition(0), FormList.Values);
+                            break;
+
+                        case "lvup":
+                            var LevelList = Kh2.Battle.Lvup.Read(stream);
+                            var moddedLevels = deserializer.Deserialize<Kh2.Battle.Lvup>(sourceText);
+                            int i = 0;
+                            foreach (Kh2.Battle.Lvup.PlayableCharacter character in moddedLevels.Characters)
```
Xeeynamo, line 337: As you are using the index `int i` here, it might be preferable to use a `for` rather than a `foreach`.

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,79 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+                            {
+                                LevelList.Characters[i].Levels = character.Levels;
```
Xeeynamo, line 339: Isn't the patching too simplistic? How can I say I just want to patch Sora's level 60 but not anything else?

Xeeynamo (changes requested):
The idea is a very good one, but there are some required changes that looks for future implementations, about consistency or some ideas to make the patching system of those files in a better way.

Also another important code change I see it's missing here are Unit Tests. It is always good to have tests to not only verify, programmatically, that everything works as expected. But also to protect us to break compatibility with existing mods in the future if we accidentally change a critical component. Please refer to `PatcherTests.cs` to add your tests to each format of 00battle and 03system.

OpenKh.Patcher/PatcherProcessor.cs
Delta-473: But then you need to build a randomiser into the mod launcher? Personally instead of using enc in the file name I would use an tag in the yaml just to prevent issues with other mod creators using enc as a file name.

OpenKh.Patcher/PatcherProcessor.cs
tommadness: This would be ideal, however Kh2.Battle.Lvup.PlayableCharacter has no identifier for what characterID this should belong to, and Kh2.Battle.Lvup.PlayableCharacter.Level has no identifier for what level# this should apply to.

moddedLevels could either deserialize into a custom type, or the above classes could be updated to have some internally identifying information. Thoughts?

OpenKh.Patcher/PatcherProcessor.cs
tommadness: Not intended, functionality has been updated.

tommadness:
All comments marked as "resolved" have been implemented and committed.

`lvup` changes are on hold for a decision about the data structure of `Kh2.Battle.Lvup`

Unit tests are in-progress.

Xeeynamo:
I saw you added unit tests. Nice! I suggest to split your test into different smaller tests, each one for every scenario. One for the `fmlv`, another one for `lvup` and so on. In this way, if one of the functionalities break it is easier to track back which one of them is causing the test to fail.

Another small issue I suggested on Discord is to rebase your branch upon the one you want to merge to. I am not sure how it will work as you previously merged some stuff in. But try to do a `git rebase upstream/feature/openkh-mods-loader` after adding `git remote add upstream git@github.com:xeeynamo/openkh.git`.

tommadness:
It seems that my previous merges have messed up the Git Timeline and the merge gets messy. It may be best for me to close this PR, re-fork, and re-push my changes. Thoughts?

Also, I'll split the various tests.

Xeeynamo:
Closed in favour of #457

PR #456, opened 2021-03-12
Upgrade to .NET 5

Xeeynamo:
.NET 5.0 and Visual Studio 2019 are mature enough to be used. It is important to upgrade to the latest version of the SDK to gain the performance, compatibility and code improvements frequently.

PS. Later in the year I plan to upgrade to .NET 6.0 too.


Truthkey (approved):
It looks like a pretty straightforward upgrade. Merge time!

PR #457, opened 2021-03-12
Add ListReplace patching method and tests

tommadness:
Cleaned up version of PR #452 . Adds functionality to the Patcher Processor that enables editing the following files using YAML:

* 00battle.bin
  * Fmlv
  * Lvup
  * Bons
* 03system.bin
  * Trsr
  * Item
* 00objentry.bin

### DIFF
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -381,6 +381,18 @@ namespace OpenKh.Patcher
                             Kh2.Battle.Bons.Write(stream.SetPosition(0), bonusList.Values);
                             break;
 
+                        case "objentry":
+                            var objEntryList = Kh2.Objentry.Read(stream).ToDictionary(x => x.ObjectId, x=>x);
+                            var serializer = new Serializer().Serialize(objEntryList);
+                            File.WriteAllText(context.GetSourceModAssetPath("objentry.yml"), serializer);
+                            var moddedObjEntry = deserializer.Deserialize<Dictionary<uint, Kh2.Objentry>>(sourceText);
+                            foreach (var objEntry in moddedObjEntry)
+                            {
+                                objEntryList[objEntry.Key] = objEntry.Value;
+                            }
+                            Kh2.Objentry.Write(stream.SetPosition(0), objEntryList.Values);
+                            break;
+
                         default:
                             break;
                     }
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -300,10 +300,10 @@ namespace OpenKh.Patcher
                     {
                         case "trsr":
                             var trsrList = Kh2.SystemData.Trsr.Read(stream).ToDictionary(x => x.Id, x => x);
-                            var moddedTrsr = deserializer.Deserialize<Dictionary<int, Kh2.SystemData.Trsr>>(sourceText);
+                            var moddedTrsr = deserializer.Deserialize<Dictionary<ushort, Kh2.SystemData.Trsr>>(sourceText);
                             foreach (var treasure in moddedTrsr)
                             {
-                                trsrList[treasure.Value.Id].ItemId = treasure.Value.ItemId;
+                                trsrList[treasure.Key].ItemId = treasure.Value.ItemId;
                             }
                                 Kh2.SystemData.Trsr.Write(stream.SetPosition(0), trsrList.Values);
                             break;
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -311,22 +311,28 @@ namespace OpenKh.Patcher
                         case "item":
                             var itemList = Kh2.SystemData.Item.Read(stream);
                             var moddedItem = deserializer.Deserialize<Kh2.SystemData.Item>(sourceText);
-                             foreach (var item in moddedItem.Items1)
-                             {
-                                var itemToUpdate = itemList.Items1.FirstOrDefault(x => x.Id == item.Id);
-                                if (itemToUpdate != null)
+                            if (moddedItem.Items1 != null)
+                            {
+                                foreach (var item in moddedItem.Items1)
                                 {
-                                    itemList.Items1[itemList.Items1.IndexOf(itemToUpdate)] = item;
+                                    var itemToUpdate = itemList.Items1.FirstOrDefault(x => x.Id == item.Id);
+                                    if (itemToUpdate != null)
+                                    {
+                                        itemList.Items1[itemList.Items1.IndexOf(itemToUpdate)] = item;
+                                    }
                                 }
-                             }
-                             foreach (var item in moddedItem.Items2)
-                             {
-                                var itemToUpdate = itemList.Items2.FirstOrDefault(x => x.Id == item.Id);
-                                if (itemToUpdate != null)
+                            }
+                            if (moddedItem.Items2 != null)
+                            {
+                                foreach (var item in moddedItem.Items2)
                                 {
-                                    itemList.Items2[itemList.Items2.IndexOf(itemToUpdate)] = item;
+                                    var itemToUpdate = itemList.Items2.FirstOrDefault(x => x.Id == item.Id);
+                                    if (itemToUpdate != null)
+                                    {
+                                        itemList.Items2[itemList.Items2.IndexOf(itemToUpdate)] = item;
+                                    }
                                 }
-                             }
+                            }
                             itemList.Write(stream.SetPosition(0));
                             break;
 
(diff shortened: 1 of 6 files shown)


tommadness:
This should now address all requested changes from @Xeeynamo outlined in the comments of PR #452.

Xeeynamo:
Hi Thomas, Can I suggest to change your FmlvDTO with this simpler contract?

```csharp
    public class FmlvDTO
    {
        public int Level { get; set; }
        public int Experience { get; set; }
        public int AbilityLevel { get; set; }
        public int Ability { get; set; }
    }
```

and ideally on your test set-up you might want to have something like the following:
image

You can populate `FormId` from the name of the form easily. You do not need to populate FormVanilla or FormFm as they are just used as helpers for `FormId`.

tommadness:
> Hi Thomas, Can I suggest to change your FmlvDTO with this simpler contract?
> 
> ```cs
>     public class FmlvDTO
>     {
>         public int Level { get; set; }
>         public int Experience { get; set; }
>         public int AbilityLevel { get; set; }
>         public int Ability { get; set; }
>     }
> ```
> 
> and ideally on your test set-up you might want to have something like the following:
> image
> 
> You can populate `FormId` from the name of the form easily. You do not need to populate FormVanilla or FormFm as they are just used as helpers for `FormId`.

Implemented. I decided to keep some slightly more descriptive field names, but removed unnecessary fields.

Xeeynamo:
Regarding the FMLV, I improve the test as it was not checking if the input and modded FMLV were merged. Be free to do a `git fetch && git cherry-pick 5feee5d19291c29861c383cd608a62c153126630` if you want to add it into your PR. The change is the following: 5feee5d19291c29861c383cd608a62c153126630

OpenKh.Tests/Patcher/PatcherTests.cs
```diff
@@ -514,6 +514,638 @@ public void MergeKh2AreaDataScriptTest()
...
+
+            patcher.Patch(AssetsInputDir, ModOutputDir, patch, ModInputDir);
+
+            AssertFileExists(ModOutputDir, "00battle.bar");
+
+            File.OpenRead(Path.Combine(ModOutputDir, "00battle.bar")).Using(stream =>
+            {
+                var binarc = Bar.Read(stream);
+                var bonsStream = Kh2.Battle.Bons.Read(binarc[0].Stream);
+                Assert.True(bonsStream[0].BonusItem1 == 200);
```
Xeeynamo, line 859: Prefer to use `Assert.Equal(200, bonsStream[0].BonusItem1)` instead

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,124 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+            {
+                    string sourceText = File.ReadAllText(context.GetSourceModAssetPath(source.Name));
+                    switch (source.Type)
+                    {
+                        case "trsr":
+                            var trsrList = Kh2.SystemData.Trsr.Read(stream).ToDictionary(x => x.Id, x => x);
+                            var moddedTrsr = deserializer.Deserialize<Dictionary<int, Kh2.SystemData.Trsr>>(sourceText);
+                            foreach (var treasure in moddedTrsr)
+                            {
+                                trsrList[treasure.Value.Id].ItemId = treasure.Value.ItemId;
```
Xeeynamo, line 306: I can see what you are doing here and why. But if in the future someone wants to effectively patch the `trsr` with the other allowed values, modifying this implementation to support it will break the existing mods. What if you check the world value? If it is `0` then patch the `ItemId`. If it is not `0` then patch everything in the record.

Xeeynamo: Also what if the Value ID (or the index of the array) is too big? You might want to add a new item to the list at that point.

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,124 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+                            }
+                                Kh2.SystemData.Trsr.Write(stream.SetPosition(0), trsrList.Values);
+                            break;
+
+                        case "item":
+                            var itemList = Kh2.SystemData.Item.Read(stream);
+                            var moddedItem = deserializer.Deserialize<Kh2.SystemData.Item>(sourceText);
+                            if (moddedItem.Items1 != null)
```
Xeeynamo, line 314: This was my bad... `Items1` should be `Items` and `Items2` should be `Stats`. Do you want to rename them? Do a CTRL+R+R to ensure you are renaming it everywhere in the codebase.

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,124 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+                            {
+                                foreach (var item in moddedItem.Items1)
+                                {
+                                    var itemToUpdate = itemList.Items1.FirstOrDefault(x => x.Id == item.Id);
```
Xeeynamo, line 318: Same here: if the item is not in the list, add it. Also is the ID ordered from `0` to `n`? If yes you might want to re-order the final list. If not, just leave it be.

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,124 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+                                    {
+                                        itemList.Items1[itemList.Items1.IndexOf(itemToUpdate)] = item;
+                                    }
+                                }
+                            }
+                            if (moddedItem.Items2 != null)
+                            {
+                                foreach (var item in moddedItem.Items2)
+                                {
+                                    var itemToUpdate = itemList.Items2.FirstOrDefault(x => x.Id == item.Id);
```
Xeeynamo, line 329: Same here, add if it does not exists

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,124 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+
+                            levelList.Write(stream.SetPosition(0));
+                            break;
+
+                        case "bons":
+                            var bonusList = Kh2.Battle.Bons.Read(stream).ToDictionary(x => String.Concat(x.RewardId, characterMap.FirstOrDefault(y => y.Value == x.CharacterId).Key), x => x);
+                            var moddedBonus = deserializer.Deserialize<Dictionary<string, Kh2.Battle.Bons>>(sourceText);
+                                foreach (var bonus in moddedBonus)
+                                {
+                                    bonusList[bonus.Key] = bonus.Value;
```
Xeeynamo, line 385: Add if not found

OpenKh.Patcher/PatcherProcessor.cs
```diff
@@ -279,5 +284,124 @@ private static void PatchAreaDataScript(Context context, List<AssetFile> sources
...
+                                }
+                            Kh2.Battle.Bons.Write(stream.SetPosition(0), bonusList.Values);
+                            break;
+
+                        case "objentry":
+                            var objEntryList = Kh2.Objentry.Read(stream).ToDictionary(x => x.ObjectId, x=>x);
+                            var moddedObjEntry = deserializer.Deserialize<Dictionary<uint, Kh2.Objentry>>(sourceText);
+                            foreach (var objEntry in moddedObjEntry)
+                            {
+                                objEntryList[objEntry.Key] = objEntry.Value;
```
Xeeynamo, line 395: Add if not found

OpenKh.Tests/Patcher/PatcherTests.cs
```diff
@@ -514,6 +514,638 @@ public void MergeKh2AreaDataScriptTest()
...
+                        Stream = trsrStream
+                    }
+                });
+            });
+
+            File.Create(Path.Combine(ModInputDir, "TrsrList.yml")).Using(stream =>
+            {
+                var writer = new StreamWriter(stream);
+                writer.WriteLine("1:");
+                writer.WriteLine("  Id: 1");
```
Xeeynamo, line 576: Is the key always equal to the Id? If that is the case it might be possible to simplify this.

OpenKh.Tests/Patcher/PatcherTests.cs
```diff
@@ -514,6 +514,638 @@ public void MergeKh2AreaDataScriptTest()
...
+                        Type = Bar.EntryType.List,
+                        Stream = bonsStream
+                    }
+                });
+            });
+
+            File.Create(Path.Combine(ModInputDir, "BonsList.yml")).Using(stream =>
+            {
+                var writer = new StreamWriter(stream);
+                writer.WriteLine("15Sora:");
```
Xeeynamo, line 844: It is not 100% clear to me why there is a `15` before the name of the character? 🤔

Xeeynamo (changes requested):

OpenKh.Tests/Patcher/PatcherTests.cs
tommadness: This was my way of creating a unique key of `BonusID` and `Character`. However, in keeping with other changes we've made, this should likely change now.

I will change this to 
```
BonusID:
- Character:
  - Reward ID:
    ...
- Character:
```

OpenKh.Patcher/PatcherProcessor.cs
tommadness: Addressed in https://github.com/Xeeynamo/OpenKh/pull/457/commits/136d8d16700419aa10cc79c44aa29c15f5fb2335

OpenKh.Tests/Patcher/PatcherTests.cs
tommadness: I assume this is true of the `Assert.True`s in the test suite?

Xeeynamo: Yes

OpenKh.Patcher/PatcherProcessor.cs
tommadness: Item list isn't ordered internally, leaving the order as-is.

tommadness:
All requested changes have been implemented.

Xeeynamo (approved):

Xeeynamo:
As the last change I require, I think that `listreplace` can be changed as `listpatch` as it is now designed to patch the existing content rather than replacing it. After taking this decision, I will be super happy to merge this 👍

PR #458, opened 2021-03-16
More Birth by Sleep Tools

Truthkey:
This PR includes the following changes:

Tools & Implementation:
- PAtk Editor (Controls attack values)
- OLO Editor (Enemy Spawns)
- Mission Editor (Rewards and cutscenes for missions)

Dictionary:
- Command List
- Ability List
- Game Objects (Identifiers with name of objects)

Improvements:
- PMO Converter fixes and vertex color support.
- PMP Writer now properly writes object offsets for PMP Converter.
- Added Final Mix folder for BBSA tool.

Other:
- Initial implementation of the SCD format.


Xeeynamo:
I get a build error. You now might want to change your project to use .NET 5 instead of .NET Core 3.1. It is very simple as doing a find and replace from `netcoreapp3.1` to `net5.0` (or `net5.0-windows` for your WinForms apps). When you do that, be sure to do a "Recompile solution"

OpenKh.Bbs/SystemData/Command.cs
```diff
@@ -6,7 +6,7 @@ namespace OpenKh.Bbs.SystemData
 {
     public class Command
     {
-        public enum Type : ushort
+        public enum Kind : ushort
         {
             COMMAND_KIND_None = 0,
```
Xeeynamo, line 11: Unlike C/C++, you could rename `COMMAND_KIND_None` to just `None` as you might need to do `Command.Kind.None` anyway to use it. Same for all the other enums you addded.

Xeeynamo (approved):
Impressive amount of work. There are very strong points about your research of new and existing file formats. Birth By Sleep modding became easier and easier thanks to you!

I left a comment of a possible request change. Let me know what you think about it.

OpenKh.Bbs/SystemData/Command.cs
Truthkey: Alright, that makes sense. I will change them whenever I get home.

PR #459, opened 2021-03-18
Kingdom Hearts 1 ISO extractor and BinArc parser

Xeeynamo:
The tool `OpenKh.Command.IdxImg` is now able to extract a Kingdom Hearts 1 ISO as well.

Also KH1 has some kind of header-less BAR used for the majority of the files in the game. I added a parser but not a tool as I need to do more research into it first.


OpenKh.Command.IdxImg/KingdomHearts1.cs
```diff
@@ -0,0 +1,118 @@
...
+                }
+            }
+
+            private class ListCommand
+            {
+                private Program Parent { get; set; }
+
+                [Required]
+                [FileExists]
+                [Argument(0, Description = "Kingdom Hearts II IDX file, paired with a IMG")]
```
Delta-473, line 91: Needs to be Kingdom Hearts I ;)

Xeeynamo: LOL thank you x'D

Truthkey (approved):
Fully working and all issues are fixed.

PR #462, opened 2021-03-24
spawnpoint method

snelson3:
Spawnpoint method for the mod manager. It would be nice to be able to provide partial yaml file and mege it into the existing spawnpoint, but I wanted to get this out first since functionally that would just be an improvement.

Example mod using this method: https://github.com/thundrio-kh/kh2-spawnpoint-example


Xeeynamo (approved):
It make perfect sense in the way you developed it. I really wanted to patch the existing file rather than replacing it with. But for how complex and dynamic the spawnpoint can be I doubt we can ever find a clean solution. I am going to approve this as it is 👍  good work

snelson3:
There were some merge conflicts caused by the listpath PR, but I fixed those and this is ready to merge now.

PR #467, opened 2021-03-28
Fix broken OpenKh.Command.IdxImg

Xeeynamo:
The PR #459 introduced some regressions where it was no longer possible to use the tool at all but to extract KH1 game's data. This restores all the intended functionalities.


Rikux3 (approved):
Tested with both KH1 and KH2 extraction. Works now!

PR #476, opened 2021-04-08
add missing europe product id kh2

snelson3:
One of the product id's for EU Vanilla KH2 was missing, preventing the mods manager from extracting it.


Xeeynamo:
I double-checked my code and I actually mistaken at writing `SLES_541.14` as `SLES_541.44`. Code must be changed also [here](https://github.com/Xeeynamo/OpenKh/blob/master/OpenKh.Tools.ModsManager/Services/Pcsx2MemoryService.cs) and in the `GameDetectorModel` you added (you can delete `SLES_541.44`)

Delta-473:
Couldn't push to this branch so generated a diff file with fixes that let it work on my machine, also I did update the UK version number as it should be the same but with different prefix?
[patchfile.txt](https://github.com/Xeeynamo/OpenKh/files/6682641/patchfile.txt)

PR #480, opened 2021-04-09
A lot of hashes and add assets inside folder

efonte:


Xeeynamo:
Please I ask to double check the changes next time. This PR not only removed some hashes I previously found, but also inserted un-necessary files in `kh2idx.txt` they were already extracted successfully.

PR #485, opened 2021-04-12
Feature/egs hed packer

Noxalus:
- [X] ~Create a new `pack`  command to repack PKG/HED files from a folder~
- [x] Create a new `patch` command to replace a list of files in a PKG
- [X] Handle files encryption. (C = A XOR B <=> A = C XOR B, so not to difficult 😅 )
- [X] Handle file compression
- [X] Handle HD assets repack
- [x] Handle new files for `patch` command


Truthkey:
This is super exciting! I'll give it a test and merge it if it works!
How does it handle files that have a hash name?

Delta-473 (commented):
Haven't tested it yet, but at first glance it looks correct. Waiting for some more testing. It would be nice to have some unit tests but can always be added later when you can create an archive file from scratch.

Vladabdf:
Is this ready to be merged?

PR #491, opened 2021-04-30
Update pawns.md

kinlyki:
found the PS4 Exclusive Keyblade's pawn ID


Delta-473:
is this pawn only available on ps4 or does the reference still exist on the pc version?

kinlyki:
The PC version still has an effect folder they use for impact stars, and the PS4 and Xbox Exclusive Keyblades' effect stars are still in the game. the pawn ID of the ps4 keyblade is the only one missing from the github

Delta-473:
Okay then it's LGTM, otherwise there was comment needed or something if it was missing in pc build.

Delta-473 (approved):
LGTM *Forgot to do it as review.*

PR #501, opened 2021-06-17
[DOC] KH BBS network packages

Delta-473:
This document isn't finished yet, as there're more packets to be added and it needs to be split up into multiple files probably, so that packets for each gamemode are grouped together as it looks like some structs have the same name but different sizes/values depending on the game mode. Validation against packet captures and implementation of a network analyser and other tools is for future pull requests.


Truthkey (changes requested):
As the [Docs Guidelines](https://github.com/Xeeynamo/OpenKh/blob/master/DOCS_GUIDELINE.md) mentions, all offsets need to be hexadecimal. So, if you change that, I don't mind approving and merging.

Delta-473:
Offsets converted to hex, do not merge yet as I need to add packets from like 4 more classes.

Truthkey (approved):

PR #504, opened 2021-07-01
ARD Update & Libretto Documentation

1234567890num:
Still a lot of unknown, but probably better than nothing

### DIFF
--- a/docs/kh2/file/type/areadata.md
+++ b/docs/kh2/file/type/areadata.md
@@ -47,6 +47,8 @@ When referring to _entrance_ it means that is the index where to spawn the chara
 | 24     | int   | Unknown
 | 28     | int   | Always 0 [*¹](#notes)
 
+Approach Direction Trigger determines if the event is triggered only when approaching from a specific direction or if it can be triggered by approaching it from any direction (even from above/below). An example of the former would be the room transition used in CoR skip where you land behind the trigger and have to move away from the exit for a bit in order to move to CoR: Depths. An example of the latter would be Cloud's mandatory cutscene in HB.
+
 #### Entity
 
 | Offset | Type  | Description
--- a/docs/kh2/file/type/areadata.md
+++ b/docs/kh2/file/type/areadata.md
@@ -346,7 +346,7 @@ Seems to recursively call the spawn script parser, but it is only used in `he09`
 
 #### Unknown1d
 
-Purpose unknown. Used 196 times.
+Auto revert Sora when the room is entered if the argument is 4 and disables drive if the argument is 5. Used 196 times.
 
 #### BattleLevel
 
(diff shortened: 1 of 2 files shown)


docs/kh2/file/type/libretto.md
```diff
@@ -0,0 +1,27 @@
+# [Kingdom Hearts II](../../index.md) - Libretto
+
+This is the file that connects the Text ID in MSG files with Talk Message in ARD files. The world-specific libretto files is a BAR file with only 1 subfile while 13libretto.bin is the subfile directly. The structure below assumes that you have extracted said subfile.
+
+## Header
+
+| Offset | Variable Type | Description |
```
Delta-473, line 7: Can you rename 'Variable Type' to 'Type' as everywhere else in the documentation it's called 'Type' also please the table layout ;)

docs/kh2/file/type/areadata.md
```diff
@@ -41,7 +41,7 @@ When referring to _entrance_ it means that is the index where to spawn the chara
 | 18     | int   | Always 0 [*¹](#notes)
 | 1c     | byte  | [Area ID of a world](../../worlds.md)
 | 1d     | byte  | Entrance
-| 1e     | byte  | [World ID](../../worlds.md)
+| 1e     | byte  | Approach Direction Trigger
```
Delta-473, line 44: Is there somewhere a description on what this is?

1234567890num: It's basically determines if the event is triggered only when approaching from a specific direction or if it can be triggered by approaching it from any direction (even from above/below). An example of the former would be the room transition used in CoR skip where you land behind the trigger and have to move away from the exit for a bit in order to move to CoR: Depths. An example of the latter would be Cloud's mandatory cutscene in HB.

The above explanation can probably be worded more concisely, but I can't find the right words.

docs/kh2/file/type/libretto.md
1234567890num: `Variable Type` has been renamed. I don't understand what you mean by the table layout.

Delta-473: afbeelding
Not a requirement but I prefer if the table column lines are straight.

docs/kh2/file/type/areadata.md
Delta-473: If you can add that somewhere as a note then maybe other people can dive deeper into it or reword it.

1234567890num: Explanation added.

docs/kh2/file/type/libretto.md
1234567890num: For some reason the preview has the table unaligned but the editor has it aligned.

Delta-473:
LGTM

Vladabdf (approved):

PR #509, opened 2021-07-25
[Test]: add net 6.0 target

Delta-473:
A branch that adds target support for .net 6.0 core. All the test runs the same as .net 5.0 but some real testing would be preferable to see if there aren't regressions. Also .net 6 is not out yet and in beta still, but this is a first step to get native Linux support for the GUI tools with the "new" MAUI GUI lib.
afbeelding

### DIFF
--- a/IteEditor/OpenKh.Tools.IteViewer.csproj
+++ b/IteEditor/OpenKh.Tools.IteViewer.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
-    <TargetFramework>net5.0-windows</TargetFramework>
+    <TargetFrameworks>net5.0-windows;net6.0-windows</TargetFrameworks>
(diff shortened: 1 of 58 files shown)


Xeeynamo:
@Delta-473  Would make sense to remove .NET 5 support (eg. from `<TargetFrameworks>net5.0-windows;net6.0-windows</TargetFrameworks>` to `<TargetFramework>net6.0-windows</TargetFrameworks>`)? .NET 6 is out and stable since a few months now and it is ready to be used.

Delta-473:
Yeah at this point I think it's fair to just move to .NET 6 as you can still use VS2019 and it gives speed improvements across the board. In OpenKH.Engine the cutscenes play a bit smoother on older CPU's. Wanted to try to benchmark it on my Raspberry Pi but there where issues initialising somewhere for the 3D part.

Xeeynamo:
Supersede by #598

Issue #506, opened 2021-07-21
[BUG] - Mod manager TryParse functions fail depending on Windows System region


Delta-473:
**OpenKH Build**
335 

**OpenKH Program**
Mod manager

**Describe the Bug**

**To Reproduce**
1. Download the randomiser mod https://kh2-randomizer.herokuapp.com/seed/MCORFBIY
2. Press build and it will fail as the Single.TryParse function fails in the Ard files (the reason for this is as the floats it uses are in this format (-3.3895395E+38) while other system regions work with (-3,3895395E+38). TLDR decimal character used issue.

**Expected behavior**
Needs to build mod.

**Additional context**
https://docs.microsoft.com/en-us/dotnet/api/system.single.tryparse?view=net-5.0

Delta-473:
"fixed" with #511

PR #512, opened 2021-08-10
[Testing needed] Make float parsing culture invariant

Delta-473:
Fixes issue #506 but the randomiser mod https://kh2-randomizer.herokuapp.com/seed still generates this error 
afbeelding
Also this specific mod issue might go away after #511 gets merged.


Truthkey:
#511 got merged. Could you please update your branch and see if the issue still exists?

Delta-473:
ParseAsFloat is unused for now so the issue is somewhere else, also the author of the randomiser is gonna need to update the tool as Capacity is an int now instead of float.

PR #524, opened 2021-10-21
Table of KH1 MDLS files w/ character names

Vladabdf:
The list of files should be credited to Rapid; he compiled the list, I just converted it to markdown format for the repo. Furthermore, I included a couple edits and a "Return to Index" clickable link to every game-specific index document so that users can actually return to the OpenKH.dev homepage instead of having to type it back into the URL bar. Idk why this wasn't a thing before, but here we are now.

It also goes without saying, sometime in the future the "Heartless" names will need to be changed to the appropriate names. E.g. Shadow (Heartless), etc., but that can be done later. For now, having the dictionary at all is a step up.

Rapid says KH1 WPN and DDD model files should be next as well.


Vladabdf:
It seems I accidentally merged this with another branch. Whoops. I'll try to fix all that.

PR #539, opened 2022-01-02
Tools/ddd rbin

thamstras:
DDD rbin extraction

based on info from: https://forum.xentax.com/viewtopic.php?f=10&t=9884&sid=7c58553d29dcad66cebddfb3689f4da7&start=15#p100644


thamstras:
Original LZovl/BLZ decompression code from: https://www.gbadev.org/tools.php?showinfo=1439

thamstras:
The code is still a bit messy and probably doesn't quite do enough error checking, but we can now extract single files, entire rbins or entire folders of rbins.

Truthkey:
> The code is still a bit messy and probably doesn't quite do enough error checking, but we can now extract single files, entire rbins or entire folders of rbins.

Are you planning to continue working on this? If you don't, I'll merge it.

thamstras:
I put in some more descriptive error messages and added some documentation. That'll probably do.

Truthkey (approved):

PR #548, opened 2022-02-01
Update pawns.md

Kenpeidan:
Added some missing entries, some of which are duplicates.


Delta-473:
Thx for contributing, you don't need to make a new pull request for every set of changes. :)

Delta-473 (approved):
LGTM

PR #557, opened 2022-02-15
Custom HD Asset Support

1234567890num:
Based on [KH PC Patch Manager](https://github.com/AntonioDePau/KHPCPatchManager) and also worked on by DA, this allows KH2.IdxImg to include custom HD assets for most filetypes when repacking.

Supported filetypes include mdlx, imd, imz, or the latter two within a bar file. Some files are not supported such as a.fm, map, or any bar file containing a pax subfile.


Delta-473:
The tests are failing, can you check locally which test needs updating and fix it pls. Also if you have added functionality that isn't covered by the existing tests please consider adding tests for those. :)

1234567890num:
Redundant with #570 .

PR #558, opened 2022-02-15
[Docs] KH3 Battle Quotes

Minty123:
Ok, this one might be a bit special, but as it's coming up every once in a while for character mods, I thought it might as well be documented here. For now the page only lists Sora.


Delta-473 (approved):
LGTM are you going to add to this pr or do I merge already?

Minty123:
For now I'll keep it as it is since transcribing and collecting the info takes a long time. Feel free to merge! :-)

PR #570, opened 2022-02-24
Support for "linking" HD assets to SD assets on PKG patch.

o0DemonBoy0o:
num made a similar PR, but I fixed up and modified a few areas.
What this does is when a new original asset is patched in of a supported type (2dd, 2ld, bar, bin, mag, mdlx, imd, imz, pax, tm2, and a.(region)) the patcher will check if a remastered directory exists before scanning the file itself to find the image offsets for correctly linking the SD asset to the remastered textures/audio.
if the remastered folder does not exist, or if the file is not of one of those types then the file is patched in as normal.


Delta-473 (changes requested):
I don't have time to go through the entire pr but you have put a copy of EgsHdAsset.cs in the root directory by accident. Can you delete this duplicate pls?

Delta-473:
Looks pretty good at first look, just one question would it be better to extract all the data classes outside of the tool so it can be used by other tools?

o0DemonBoy0o:
Whoops, didn't notice the duplicate. that should be fixed now.
Maybe, but i'm not sure what use the data classes have for other tools. To be honest i'm also not actually sure how to even do that, my coding ability isn't actually all too great.

Delta-473:
Hmm when needed it can always been cleaned up it's not that we need to be able to display hd assets in map editor at this point.

Delta-473 (approved):
Didn't test myself but test pass so LGTM

PR #572, opened 2022-02-25
Fixed some PAX remastered texture linking issues

o0DemonBoy0o:
Thanks to Topaz reporting some bugs, I fixed how many textures in a PAX should actually count for HD linking as well as correcting the offsets for some textures where duplicates exist.
There still seems to be the issue of how you should order your remastered textures if adding a new PAX to the game as there seems to be no way to tell by reading the new file how the textures should exactly be ordered. 
So while ass the offsets in the new file would be correct, the order of the textures in the remastered folder won't always be and thus require manually fixing every now and then.

### DIFF
--- a/OpenKh.Egs/EgsTools.cs
+++ b/OpenKh.Egs/EgsTools.cs
@@ -752,6 +752,7 @@ namespace OpenKh.Egs
     class PAX
     {
         public List<int> Offsets = new List<int>();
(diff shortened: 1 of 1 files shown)


Vladabdf (approved):
Beautiful. Would you say this is ready to merge?

o0DemonBoy0o:
I would say so. After quite a bit of texting it various other types of files containing PAX I can't say I found any other bugs.

PR #573, opened 2022-02-26
Add comment to IMGD header

Vladabdf:
Clarified how to calculate the desired value, as it may not be intrinsically clear.


Delta-473 (approved):
LGTM

PR #574, opened 2022-02-27
Raw extracting and file patching

o0DemonBoy0o:
Thanks to the work of @AntonioDePau, this addition adds the ability to extract and patch files in their raw format.
A "raw" file is the header used in the PKG + the sd file and any remastered assets linked to it all bundled together.
The use case of this is pretty situational but it can allow for quicker extraction and rebuilding of entire PKGs, the ability to copy & paste a file from one PKG to another, or even pre-modify a file and extract it as a raw file for potentially smaller mod sizes.
To extract files in their raw format you just need to use the -r | --raw command line option like so:

OpenKh.Command.IdxImg.exe hed extract -r file.hed

Patching is not much different. as long as the raw file's root folder is named "raw" and the file path matches the original sd file path then it will patch fine.

This was built upon the code used for PR #572 so I think that should be looked at first before considering this one.
I submitted this as a separate PR should this idea be denied.


Vladabdf (approved):
Works wonderfully. Upon approval of #572, this will promptly be merged so long as you think it's ready.

o0DemonBoy0o:
I believe it is.

PR #577, opened 2022-02-28
Added Xemnas

Minty123:
Xemnas transcription by VaultLink. + fixed a line by TX

### DIFF
--- a/docs/kh3/battlequotes.md
+++ b/docs/kh3/battlequotes.md
@@ -275,7 +275,7 @@ Their type is SQEXSEADSoundBank or SQEXSEADSound.
 | 019       | "Begone"                         |              |
 | 020       | \<strike sounds\>                |              |
 | 021       | \<strike sounds\>                |              |
-| 022       |                                  | Audio fails to extract. File uses dummy codec. |
+| 022       | ?                                | Fails to extract. |
 | 023       | "No Escape"                      |              |
 | 024       | "Leaving?"                       |              |
 | 025       | "No"                             |              |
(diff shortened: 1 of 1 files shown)


docs/kh3/battlequotes.md
```diff
@@ -245,3 +246,38 @@ Their type is SQEXSEADSoundBank or SQEXSEADSound.
...
+| 013       | \<hurt sounds\>                  | Damage taken |
+| 014       | \<hurt sounds\>                  | Damage taken |
+| 015       | \<hurt sounds\>                  | Damage taken |
+| 016       | "Are you Ready"                  |              |
+| 017       | "Grovel"                         |              |
+| 018       | "Pay Attention"                  |              |
+| 019       | "Begone"                         |              |
+| 020       | \<strike sounds\>                |              |
+| 021       | \<strike sounds\>                |              |
+| 022       |                                  | failed to extract |
```
Vladabdf, line 278: For line 22, could we change the comment to "?" instead? If only for the sake of being completely clear that Xemnas does not, in fact, say "failed to extract". 😆 
Otherwise, thanks for the addition. 👍

Vladabdf (approved):

Minty123:
@Vladabdf, what you think about "Audio fails to extract. File uses dummy codec."? IMO that's nothing Xemnas would say and it would be clearer why the line is empty. :-)

Vladabdf:
In virtually every other file where a value is unknown for some reason or another,  if it's plaintext,  we've used "?" thus far, so I kindly ask we keep it that way for consistency.

Minty123:
Sorry, my bad! Think I understood it wrongly at first. In case you can change it yourself, feel free to do so. Otherwise I'll change it later on. 👍

PR #578, opened 2022-03-01
Reorganized KH2 Index

Tyniscus:
The index page for KH2 has been rewritten as well as the following:
- Rewritten "File Types" tables to include types previously only found in the index
- Added previously hidden "icons" dictionary
- Added previously hidden "unused files" to the index


Vladabdf (approved):
Nice organization of the index page! Thank you very much.

PR #581, opened 2022-03-04
Support for HD port exclusive Icons in KH2 Text Editor.

o0DemonBoy0o:
I added all the new icon IDs added by the HD port and gave them names.
Since these don't exist in the fontimage.bar, even in the HD port, I opted to have the gui text editor display the question mark icon for them instead as a workaround.


Vladabdf (approved):
Fantastic

PR #583, opened 2022-03-06
Updates for KH2FM HD message icons

o0DemonBoy0o:
Some minor changes to the names chosen for the message icons used in the HD collection and an updated doc with info about them and their names.


Vladabdf (commented):
By chance, is the PC Left Analog icon meant to be named "form(PC)"? There are some other examples, such as The Xbox L3 button  being named "weapon-keyblade-equip(PC)". If so, I'm already in the process of making a follow-up pull request that adds some information to the text editor doc, and I will gladly fix that as well since I'm also going to organize the icons.

o0DemonBoy0o:
The PC port reassigns what some of the old PS2 button prompt and unused icons were into new icons for the new control methods. So what would be the icon for "form" for the PS2 game was repurposed into the Xbox Left analog icon in the PC port. This is why I named the image "form(PC).png". In the text editor would would use name "form" if you wanted to use the xbox left analog icon on PC. 
It's a little bit confusing I admit, though generally you would want to be using the "dynamic" icons for button icons in text anyway.

Vladabdf (approved):
Looking great. Thanks as usual!

PR #595, opened 2022-04-07
MDLX & MSET editors V1.0

osdanova:
Tools to edit MDLX and MSET files. These are basic versions and are to be improved upon in the future.

What they can do:
Edit mesh flags (Eg: turn on/off transparency/glow)
View textures and texture animations
Edit texture animation data (Eg: which texture the animation is applied to and where, animation scripts...)
Edit collision data
Save as MDLX
Edit animation bones
Edit motion triggers
Possibly crash the tool if you enter stupid data

What can't do:
Export/Import models, textures, texture animations or animations
Add/Remove data (Eg: add a texture or a new collision)


Delta-473 (approved):
LGTM

PR #268, opened 2020-11-06
Feature Addition: Support for Turkish Font Table on KH2TextEditor

Topaz-Reality:
Made a font table which gets rid of 6 characters which server no purpose in the Turkish Alphabet to make room for 3 important letters and their lowercase variants. The tool is stable with this addition.

The font table in question will be used in my long anticipated Turkish Translation of the game and this feature will make the game MUCH easier to translate.
image

PR #596, opened 2022-04-07
Updating and placing the accent ~ (ã, ã, Ã, Õ) to make the tool compatible with keyboards ABNT Portugues Brasil

HeroRickyGAMES:
Updating and placing the accent ~ (ã, ã, Ã, Õ) to make the tool compatible with keyboards ABNT Portugues Brasil


Vladabdf:
I will never merge this faulty edit no matter how many times you attempt to make it happen.

Make a request more like #268 if you want to create a Brazilian Portuguese message decoder. I will absolutely not allow you to destroy the international decoder system when the accented characters you're trying to include don't even exist in the official and unmodified KH2 font table.

PR #598, opened 2022-04-17
upgrade .Net version to 6.0

Delta-473:
Replaces #509 
Target 5.0 with 6.0 sdk
afbeelding
Target 6.0 with 6.0 sdk
afbeelding

There were some people in some regions where VS2022 doesn't run well on older hardware and want to still use VS2019 instead. However this isn't possible with the release version of the 6.0 SDK. However you can use the preview 4 build of the SDK with VS2019 as a stop gap solution. I don't know How long that would work if some new features are being used.

### DIFF
--- a/IteEditor/OpenKh.Tools.IteViewer.csproj
+++ b/IteEditor/OpenKh.Tools.IteViewer.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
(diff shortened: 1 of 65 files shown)


Xeeynamo (approved):
I do not see anything wrong with this PR. The performance impact is shocking too! I approve the code change 👍 .

PS. What about that failing test? How did it penetrated to the master branch?

Delta-473:
Idk how long that test is broken. That one should be fixed in a follow up PR.

PR #599, opened 2022-04-24
Mset update - Motion transfer, Frame edition and Docs update

osdanova:
Mset editor updated:
The Interpolated Motion file is fully parsed, it can be serialized and deserialized in a lossless way.
Added edition for frame related data. Other data has not been made available due to its complexity. The file is fully parsed so if anyone needs to add something to the editor it's just a matter of adding the UI. At this point, a parser to a commonly used animation format would be better (Same with models)
An export motion to JSON and motion have been added in case someone wants to see the data parsed in a readable way.

A motion transfer between msets have been added:
- Change mode to transfer (Bottom left)
- Drag and drop the msets to the from/to mset sections.
- Bone count of the receiving mset is necessary. If the mset had valid ANBs it will be detected on file load.
- Bone transfers are required. The bone list is loaded when an ANB from the donnor mset is selected.
- The user can either add the bone transfers manually or drag and drop a CSV file with the table.
- Select the donnor and receiver MSET slots and click transfer. Lastly save the file as MSET.

Docs have been updated with all of the data from the decomp and my research.

CSV file to test Sora to Roxas transfer: [Sora2Roxas.csv](https://github.com/Xeeynamo/OpenKh/files/8549553/Sora2Roxas.csv)


Delta-473 (approved):
Didn't test but LGTM

PR #607, opened 2022-05-30
Fix PNG decompression

Xeeynamo:
The new .NET 6 #598 has a breaking change on how [GZip decompression is handled](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/partial-byte-reads-in-streams). We are using the library `DeflateStream` which is mainly used by `PngImage`. This bug affects the following products:

* PmoConverter
* PmpConverter
* Mods Manager
* Image Viewer

Essentially it is now impossible to just read byte by byte from `DeflateStream` so now we need a buffer to unload the decompressed data before reading it from a stream. Luckily `BufferedStream` does the trick and the fix is quite straightforward. I also took the opportunity to improve the error logging if a PNG acts weird.

### DIFF
--- a/OpenKh.Imaging/PngImage.cs
+++ b/OpenKh.Imaging/PngImage.cs
@@ -273,7 +273,7 @@ namespace OpenKh.Imaging
             }
             else
             {
-                throw new NotSupportedException();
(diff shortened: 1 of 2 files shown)


Delta-473 (approved):
LGTM

PR #608, opened 2022-05-30
Enhance Mods Manager injection capabilities for PSP/PS2 games

Xeeynamo:
I found that the MIPS instructions I wrote long time ago were either incomplete or incorrect. This code patch expands the capability of writing and injecting custom MIPS code to any MIPS Little Endian's code (eg. PlayStation 2 and PlayStation Portable).


Delta-473 (approved):
I hope non of the game patches breaks but LGTM

PR #609, opened 2022-05-31
PngImage fix for .NET 6

Xeeynamo:
#607 did not fix it for all the scenario. This implementation is a bit inefficient but should do the trick...


Truthkey (approved):
Awesome. A very needed fix.

PR #614, opened 2022-06-25
Adding some corrections about the SHOP file.

Dee-Ayy:
I recently took apart and learned a bit more about the SHOP format so I added what I learned and corrected some bits of info.


Delta-473 (approved):
LGTM

PR #617, opened 2022-07-07
A Fix for Relative Administrator Path Bug

### LINKED ISSUE #616: OpenKh Mod Manager Creating 'mods' folder in system32
**OpenKH Build**
1878

**OpenKH Program**
OpenKh.Tools.ModsManager.exe

**Describe the Bug**
Mod Manager is creating a `mods` folder in the users `system32` folder when installing mods from GitHub. 

**To Reproduce**
Not sure how to reproduce. We have seen it happen to a few people over the last couple of months. We close out of the setup wizard when installing mods for rando, but I have attempted to re-produce the issue myself and could not. 

**Expected behavior**
Extract fresh openkh folder to desktop, run OpenKH.Tools.ModsManager.exe, close out of the wizard, click `mods`, `install new mod`, add a mod from github. It will create and `mods` folder in System32 and download to there. 

**Screenshots**
Openkh-System32

**Additional context**
Worked with a user who had it happen during original setup. Had them download a fresh version of openkh and try to install a mod through GitHub. It still downloaded the mod to their system32 folder.

TopazTK:
Fixes #616.

### DIFF
--- a/OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
+++ b/OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
@@ -51,8 +51,9 @@ namespace OpenKh.Tools.ModsManager.Services
             }
         }
 
-        private const string ConfigPath = "mods-manager.yml";
-        private const string EnabledModsPath = "mods.txt";
+        private static string _assString = Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location);
+        private static string ConfigPath = _assString + "\\mods-manager.yml";
+        private static string EnabledModsPath = _assString + "\\mods.txt";
         private static readonly Config _config = Config.Open(ConfigPath);
 
         static ConfigurationService()
@@ -104,7 +105,7 @@ namespace OpenKh.Tools.ModsManager.Services
 
         public static string ModCollectionPath
         {
-            get => _config.ModCollectionPath ?? Path.GetFullPath("mods");
+            get => _config.ModCollectionPath ?? Path.GetFullPath(_assString + "\\mods");
             set
             {
                 _config.ModCollectionPath = value;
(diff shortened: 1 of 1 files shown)


OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
```diff
@@ -51,8 +51,9 @@ public static Config Open(string fileName)
             }
         }
 
-        private const string ConfigPath = "mods-manager.yml";
-        private const string EnabledModsPath = "mods.txt";
+        private static string _assString = Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location);
```
Xeeynamo, line 54: Could you rename it to something more meaningful? Like `AppDirectory`, `AppPath` or similar

OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
```diff
@@ -51,8 +51,9 @@ public static Config Open(string fileName)
...
+        private static string ConfigPath = _assString + "\\mods-manager.yml";
```
Xeeynamo, line 55: I highly suggest to use `Path.Combine` rather than `+ "\\"` as it internally covers weird scenario that might occur. Same for the lines below.

Xeeynamo (changes requested):
I left some comments about potential changes. But overalls it looks good 👍

OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
TopazTK: :(

OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
TopazTK: Can do.

Dee-Ayy:
Does the name need to be so childish?

PR #618, opened 2022-07-07
Do not use end_of_line from .editorconfig

Xeeynamo:
Fixes this damn message:

image

The current `.editorconfig` forces to use a certain end of line format. But on Windows and Linux, git transforms the end of line format based to the target operating system and Visual Studio follows that logic. The existing `.editorconfig` clashes with it making that issue.

This PR fixes the problem once for all for any operating system and for any IDE and text editor.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,23 +2,12 @@ root = true
 
 [*]
 charset = utf-8
-end_of_line = lf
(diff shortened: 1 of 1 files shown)


Delta-473 (approved):
Best change ever :D

PR #619, opened 2022-07-07
Fix OpenKH.Engine crash

Xeeynamo:
When closing the engine there was an occasional crash. This is now fixed. I also re-enabled the entity animation as it was commented out for some obscure reason.


Delta-473 (approved):
LGTM

PR #625, opened 2022-07-17
Feature/openkh mods loader egs

MainMemory:
The mod loader works now; somewhat. It can replace most files in KH2 (I haven't even begun trying to support other games), including the remastered assets. It doesn't support streamed music (title/gummi/end) or videos. It allows for DLL files to be loaded in order to run custom code within the game (I converted the lua file for the Garden of Assemblage mod into C++ as a test).

PR #627, opened 2022-07-17
BFMM Functionality in Mod Manager.

TopazTK:
Build_From_MM was a Python program used which allowed for Mod Manager mods to be used in the PC ports.
This PR basically takes in everything it can do and implements it to Mod Manager itself so that **we don't have to fiddle around.**

What it does:
- Allowing the Mod Manager to be set up for the PC ports.
- Building for OpenKH/PS2/PC specifically (Depending on what was declared in the wizard.)
- Extraction, Patching [Regular and "Fast"], and Backup of the EGS games.
- Restoration of the backups made with MM.

What it adds to "mod.yml":
- The "platform" tag to specifically state if a file is for a specific platform [pc, ps2, both] (Assumed "both" if not present.)
- The "package" tag to use with PC only. Tells a file where exactly to be patched to [kh2_* first, second, third, fourth, fifth, sixth] (Ignored on non-PC platforms. Assumed "kh2_first" if not present.)

From the tests made, it works surprisingly well. And it is able to patch the game ridiculously fast and is compliant with everything OpenKH tries to accomplish. Some of the "OpenKh.Egs" functions were implemented to MM separately as trying to use them would mean there would be no indication as to what was done.

I know the code is kinda messy, and there are still some things I have to do, I have to tidy up the code, add kh2pcpatch support, make sure shit works with everything etc. But this update was heavily requested by the majority of the modding community, and I had to deliver.

### DIFF
--- a/OpenKh.Egs/EgsTools.cs
+++ b/OpenKh.Egs/EgsTools.cs
@@ -18,8 +18,6 @@ namespace OpenKh.Egs
         private const string ORIGINAL_FILES_FOLDER_NAME = "original";
         private const string REMASTERED_FILES_FOLDER_NAME = "remastered";
 
-        delegate void ActionRef<T1, T2, T3, T4, T5>(T1 item1, T2 item2, T3 item3, ref T4 item4, T5 item5);
-
         #region MD5 names
 
         private static readonly IEnumerable<string> KH2Names = IdxName.Names
--- a/OpenKh.Egs/EgsTools.cs
+++ b/OpenKh.Egs/EgsTools.cs
@@ -18,6 +18,8 @@ namespace OpenKh.Egs
         private const string ORIGINAL_FILES_FOLDER_NAME = "original";
         private const string REMASTERED_FILES_FOLDER_NAME = "remastered";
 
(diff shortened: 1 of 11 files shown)


TopazTK:
Closing due to the merge of #625, which makes this PR broken and in minimal necessity.
The YAML edits make to this can be done to #625 with minimal effort should they be necessary.

MainMemory:
I strongly disagree with this

TopazTK:
Reopened due to immense backlash. <3

shananas:
This PR would at least help the Rando community a lot as we already have to use a second program to patch our seeds. Yes the mod loader would be preferred but this implements extraction in ModsManager and the ability to make 1 mod for both PC and Emulator with the tag feature. Dee-Ayy already did a good writeup on the Mod Manager Improvements about why tags are needed because of region and file differences between PC and Emulator so ill add 1 thing.
If you want to use the `type: internal` method for `a.fm` or `a.us` files the mod will only be compatible with emulator or PC respectively. being able to tell MM ignore a.fm/a.us so the build does not fail allows for provided less vanilla files and use of `type: internal` in 1 mod for both PC an emulator

TopazTK:
Closing due to patching "not fitting into OpenKH", not to be opened again.

PR #633, opened 2022-07-20
Make Panacea Game/Version Agnostic

MainMemory:
It now uses signatures to locate the functions it needs to hook, and my initial test shows that the signatures get a hit in every game in the 1.5+2.5 collection, including the launcher and theater. Currently I've left it locked to only supporting KH2, until such a time as the mod manager can support multiple games.


Delta-473 (approved):
LGTM

PR #636, opened 2022-07-21
Improve the Mod Manager for PC Modding Experience.

TopazTK:
_"Yes, patching is outside of the scope of OpenKH, but discrimination is too, so it will be merged as long as it works and is received well. And we prioritize function, user accessibility, and friendliness even above our functionality standards."_

After some private discussion with some of the other collaborators, I decided to re-open this PR from scratch to hopefully finally make it in. I will just copy and paste my statement from the PR prior:

Build_From_MM was a Python program used which allowed for Mod Manager mods to be used in the PC ports.
This PR basically takes in everything it can do and implements it to Mod Manager itself so that we don't have to fiddle around with some other executable and basically have the Mod Manager do everything on it's own.

What it does:

- Allowing the Mod Manager to be set up for the PC ports.
- Building for OpenKH/PS2/PC specifically (Depending on what was declared in the wizard.)
- Extraction, Patching [Regular and "Fast"], and Backup of the EGS games.
- Restoration of the backups made with MM.

What it adds to "mod.yml":

- The "platform" tag to specifically state if a file is for a specific platform [pc, ps2, both] (Assumed "both" if not present.)
- The "package" tag to use with PC only. Tells a file where exactly to be patched to [kh2_* first, second, third, fourth, fifth, sixth] (Ignored on non-PC platforms. Assumed "kh2_first" if not present.)

From the tests made, it works surprisingly well. And it is able to patch the game ridiculously fast and is compliant with everything OpenKH tries to accomplish. Some of the "OpenKh.Egs" functions were implemented to MM separately as trying to use them would mean there would be no indication as to what was done.

I know the code is kinda messy, and there are still some things I have to do, I have to tidy up the code, add kh2pcpatch support, make sure shit works with everything etc. But this update was heavily requested by the majority of the modding community, and I had to deliver.

Should this PR get merged, there will be the addition of the Mod Loader by MainMemory (which currently resides on another branch) and some tweaks to make MM compatible with all the other games.

### DIFF
--- a/OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
+++ b/OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
@@ -4,6 +4,7 @@ using OpenKh.Tools.ModsManager.Models;
 using OpenKh.Tools.ModsManager.Services;
 using OpenKh.Tools.ModsManager.Views;
 using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.IO;
@@ -35,6 +36,10 @@ namespace OpenKh.Tools.ModsManager.ViewModels
         private Process _runningProcess;
         private bool _isBuilding;
 
+        private const string RAW_FILES_FOLDER_NAME = "raw";
+        private const string ORIGINAL_FILES_FOLDER_NAME = "original";
+        private const string REMASTERED_FILES_FOLDER_NAME = "remastered";
+
         public string Title => ApplicationName;
         public string CurrentVersion => ApplicationVersion;
         public ObservableCollection<ModViewModel> ModsList { get; set; }
@@ -45,6 +50,8 @@ namespace OpenKh.Tools.ModsManager.ViewModels
         public RelayCommand MoveUp { get; set; }
         public RelayCommand MoveDown { get; set; }
         public RelayCommand BuildCommand { get; set; }
+        public RelayCommand PatchCommand { get; set; }
+        public RelayCommand RestoreCommand { get; set; }
         public RelayCommand RunCommand { get; set; }
         public RelayCommand BuildAndRunCommand { get; set; }
         public RelayCommand StopRunningInstanceCommand { get; set; }
@@ -183,20 +190,36 @@ namespace OpenKh.Tools.ModsManager.ViewModels
             BuildCommand = new RelayCommand(async _ =>
             {
                 ResetLogWindow();
-                await BuildPatches();
+                await BuildPatches(false);
                 CloseAllWindows();
             }, _ => !IsBuilding);
+
+            PatchCommand = new RelayCommand(async (fastMode) =>
+            {
+                ResetLogWindow();
+                await BuildPatches(Convert.ToBoolean(fastMode));
+                await PatchGame(Convert.ToBoolean(fastMode));
+                CloseAllWindows();
+            }, _ => !IsBuilding);
+
             RunCommand = new RelayCommand(async _ =>
             {
                 CloseRunningProcess();
                 ResetLogWindow();
                 await RunGame();
             });
+
+            RestoreCommand = new RelayCommand(async _ =>
+            {
+                ResetLogWindow();
+                await RestoreGame();
+                CloseAllWindows();
+            });
             BuildAndRunCommand = new RelayCommand(async _ =>
             {
                 CloseRunningProcess();
                 ResetLogWindow();
-                if (await BuildPatches())
+                if (await BuildPatches(false))
                     await RunGame();
             }, _ => !IsBuilding);
             StopRunningInstanceCommand = new RelayCommand(_ =>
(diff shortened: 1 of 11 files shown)


OpenKh.Tools.ModsManager/ViewModels/SetupWizardViewModel.cs
```diff
@@ -172,81 +181,194 @@ public SetupWizardViewModel()
                 FileDialog.OnOpen(fileName => OpenKhGameEngineLocation = fileName, _openkhGeFilter));
             SelectPcsx2Command = new RelayCommand(_ =>
                 FileDialog.OnOpen(fileName => Pcsx2Location = fileName, _pcsx2Filter));
+            SelectPcReleaseCommand = new RelayCommand(_ =>
```
paulblessing, line 184: In the mod loader / Panacea branch, this uses

```cs
SelectPcReleaseCommand = new RelayCommand(_ =>
                FileDialog.OnFolder(path => PcReleaseLocation = path));
```

which I think would be preferred to avoid confusion for users knowing which executable they need to choose. Should also make integrating the other branch easier if this is using the same approach as that one.

OpenKh.Tools.ModsManager/Views/InstallModView.xaml.cs
```diff
@@ -13,7 +13,8 @@ public partial class InstallModView : Window
     {
         private static readonly IEnumerable<FileDialogFilter> _zipFilter = FileDialogFilterComposer
             .Compose()
-            .AddExtensions("OpenKH mod as ZIP file", "zip");
+            .AddExtensions("OpenKH ZIP Mod", "zip")
```
paulblessing, line 16: Might consider something like

```cs
.AddExtensions("Mod Formats", "zip", "kh2pcpatch");
```

so users don't have to know to use the chooser to switch between them

OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
```diff
@@ -401,6 +424,151 @@ private void MoveSelectedModUp()
             ModEnableStateChanged();
         }
 
+        private async Task PatchGame(bool fastMode)
+        {
+            await Task.Run(() =>
+            {
+                if (ConfigurationService.GameEdition == 2)
```
paulblessing, line 431: Do we have constants defined anywhere for the game editions? If not, should we?

paulblessing:
- Tested a fresh OpenKH install and extracted the game data, everything looked good there
- Tested an existing OpenKH install with an existing game extraction, running the wizard it picked up on that existing location which was good
- Main thing I noticed was that even when full patching, most/all files for existing mods ended up in `kh2_first`. I'd have expected them to get spread around more when full patching.

OpenKh.Tools.ModsManager/Views/InstallModView.xaml.cs
TopazTK: This is intentional to a degree. Will not be altering it as of now.

OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
TopazTK: I'd rather not fix what works.

TopazTK:
> - Tested a fresh OpenKH install and extracted the game data, everything looked good there
> - Tested updating this build on top of an existing OpenKH install with an existing game extraction, running the wizard it picked up on that existing location which was good
> - Main thing I noticed was that even when full patching, most/all files for existing mods ended up in `kh2_first`. I'd have expected them to get spread around more when full patching.

I cannot yet determine where everything should go. Full patching will follow the package tags on the YAML for patching purposes. Will fix soon.

OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
paulblessing: Not suggesting it's broken, but magic numbers and strings in a codebase make it really tough for newcomers to figure out what means what. I'm not a core maintainer of this project though, so it's just a suggestion from me.

paulblessing: Looks like the precedent in the codebase is to just use the magic numbers for this. Consider an overall suggestion for future if anyone ever wants to improve the readability for newcomers in the future.

OpenKh.Tools.ModsManager/Views/InstallModView.xaml.cs
paulblessing: Reasons? I thought this was broken at first until I saw the chooser, and I'm going to be more tech savvy than most players.

OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
Delta-473: An enumeration in the future would be nice so mistakes are less common.

Delta-473 (approved):

PR #645, opened 2022-08-08
Kh2 MDLX export to FBX and Assimp utils

osdanova:
Export functionality to FBX for the MDLX editor. It uses Assimp, so it should work with other file formats, I've only tested with DAE and it requires correct offset matrices (I think), which I'm not calculating right so that's broken right now.
Added Assimp utils library to make use of Assimp easier and have a basis on how to parse models with Assimp.


kenjiuno:
It looks fine, good work. Is this ready for merging? If you have some undone tasks, you can continue before merging.

osdanova:
Yes. That's as far as I wanted to take it. I have some other MDLX changes but I'd rather have them in another PR.
I didn't notice the build step wasn't working though. I don't know how to fix that.

kenjiuno:
The build status will be updated (fixed) on next time GitHub Actions build.

GitHub says that there is a merge conflict on `OpenKh.sln` file. Could you fix it?

PR #653, opened 2022-08-22
PC mod manager small fixes and updates after initial user feedback

paulblessing:


TopazTK (approved):
Approved. But I want one more collaborator to do so.

Delta-473 (approved):

PR #655, opened 2022-08-25
Fix MapGen, adding AnbMaker, making minor changes to DoctChanger and CoctChanger. And so on

### LINKED ISSUE #220: Mapgen doesn't optimize DOCT enough.
Maps still lag even after the DOCT optimization. It isn't to do with the collision, which is what I thought it was at first. I would suggest looking at how many polygons/vif packets an official mesh groups or DOCT file can contain. I'm still convinced this is the problem with the transparency flag, because if a more simple map is loaded, or if more simple models are loaded, transparency works fine.
Edit: It would also be an incredibly welcome feature to specify the different collision types, player collision, camera collision, and light collision in the .yml file. Using a separate FBX for each based on its name.

kenjiuno:
# Changes

Fix clip material won't generate clip plane.

- The mesh has material option using `nodraw: true` was ignored by COCT generator.
 
Separate doct builder from coct builder.
vifPacketRenderingGroup is now built by doct.

Fix #220 

- COCT generator and DOCT generator won't share artifacts.

Set OctalTreeCount which was uninitialized.

- DOCT count was set to zero in map file, and it was inactive.

Tweak log message.

add multi nodes support to doct

- Currently BSP will only have 2 child nodes. This is change to supports up to 8 child nodes in the future.

fix collision builder bug, display collision stat

- Fix bbox computer bug.

Adding `nococt` and `nodoct` options.

- https://github.com/kenjiuno/OpenKh/blob/fix-mapgen/docs/tool/CLI.MapGen/index.md
- Useful to generate `SK` model

Adding OpenKh.Command.MapGenUtils

- `OpenKh.Command.MapGenUtils gen-schema .` will generate `mapdef.schema.json`.
- Visual Studio Code with this plugin [redhat-developer/vscode-yaml: YAML support for VS Code with built-in kubernetes syntax support](https://github.com/redhat-developer/vscode-yaml) will recognize schema for `mapdef.yml` file if we can tweak VSCode workspace settings well.

```.vscode/settings.json
{
    "yaml.schemas": {
      "mapdef.schema.json": [
        "mapdef.yml"
      ]
    }
}
```

`nodoct: true` will generate doct having single node with (-32767, -32767, -32767) to (32767, 32767, 32767) bounding box.

BigMesh class has been decomposed and disposed, because BigMesh has problem. BigMesh is not suitable design to get splitted models based on 3D coord. This makes COCT / DOCT generator very ineffective.

- SingleFace will be minimum unit of loaded user model.
- MaterialContainer contains list of MaterialDefs for texture generation.
- MapVifPacketBuilder will have its own domain model, without using external model.
- private MicroMesh, MicroVert, and MicroMeshMaker classes are added in MapBuilder. It is a splittable mesh representation suitable for VIF packet generation.

ISpatialNodeCutter is exchangable interface for BSPNodeSplitter.

BSPNodeSplitter will work better on split. Supporting to cut up to 8 nodes.

OpenKh.Command.DoctChanger supports `expose-map-doct` command.

OpenKh.Command.AnbMaker is added.

OpenKh.Command.MapGen generates map which fits to PC version.

CoctChanger and DoctChanger have subcommand to expose metrics in tree form:
[lk04 CollisionOctalTree dot svg](https://user-images.githubusercontent.com/5955540/197333573-8f907034-671b-4d4a-ba97-466ef09c8a92.jpg)
[lk04 DrawOctalTree dot svg](https://user-images.githubusercontent.com/5955540/197333627-5cc904ba-4123-4a29-8d9c-3c339b1028ec.jpg)


kenjiuno:
(splitting to small pieces)

PR #670, opened 2022-09-10
Transfer XeEngine.Tools.Public repository reference to OpenKh

kenjiuno:
OpenKh depends on projects from `Xeeynamo/XeEngine.Tools.Public` repository.
But OpenKh contributors cannot provide fix problems or improve functionality easily (simply it is not under control of OpenKh project).
Thus I have forked `XeEngine.Tools.Public` repository to OpenKh organization, so that we can manage it demands to our needs.
It is MIT licensed.
https://github.com/OpenKH/XeEngine.Tools.Public

This will help to fix issue like #132

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "XeEngine.Tools.Public"]
 	path = XeEngine.Tools.Public
-	url = https://github.com/Xeeynamo/XeEngine.Tools.Public
+	url = https://github.com/OpenKH/XeEngine.Tools.Public.git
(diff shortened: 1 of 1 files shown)


Delta-473:
This looks ok for me, as long as we try to upstream improvements from time to time. :)

Vladabdf:
Awesome idea. 👍

kenjiuno:
Thanks for comments, I will merge this by using administration privilege, in few days, if there is no objection with this.

kenjiuno:
Merged

PR #672, opened 2022-09-15
Seperate the Mod Loader and Patching Menus

shananas:
Also hide a menu if it cannot be used. If a setting is updated by running the wizard a restart of the Mod Manager is required to properly hide or display a menu if the changed setting means one of the two menus visibility should change.

I had added some code to the end of the wizard relay command so whenever the wizard was run the variables were properly updated but without refreshing/redrawing? the Mod Manager window the visibility of the Mod Loader or PC Menu wouldnt change without restarting Mod Manager entirely
```
if (ConfigurationService.GameEdition == 2)
{
   PC = true;
}
else
{
   PC = false;
}
```

### DIFF
--- a/OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
+++ b/OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
@@ -34,8 +34,8 @@ namespace OpenKh.Tools.ModsManager.Services
             public string Pcsx2Location { get; internal set; }
             public string PcReleaseLocation { get; internal set; }
             public int RegionId { get; internal set; }
-            public bool BypassLauncher { get; internal set; }
             public string EpicGamesUserID { get; internal set; }
+            public bool PanaceaInstalled { get; internal set; }
 
             public void Save(string fileName)
             {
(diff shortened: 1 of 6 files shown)


shananas:
kenjiuno has informed me on how to fix the visibility update issue so i have added code to the end of the setup wizard method to set the PC and Bypass variables so the visibility properly updates

kenjiuno (approved):

Vladabdf:
Brilliant change. Thanks for the contribution!

shananas:
Went ahead and pushed the Dev View Checkbox but it does not remain checked on restart.
Also added a code optimization TopazTK was holding off on for mod-manager.yml reference path due to my PR
From
```
private static string StoragePath = Directory.GetCurrentDirectory().IndexOf("system32") >= 0 ?
    Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location) :
    Directory.GetCurrentDirectory();
```
to
```
private static string StoragePath = Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location);
```

kenjiuno:
_Note:_ In this pull request, `PC` is hidden by default. It needs to turn on `Info` → `Dev View` in order to access `PC` drop down menu.

2022-09-18_13h17_02

2022-09-18_13h17_14

shananas:
Its only hidden if you have panacea installed. it assums the end user plans on either using patching or mod loader so hides one menu accordingly

kenjiuno:
Ok sorry this is my mistake 👍

PR #673, opened 2022-09-16
[KH2SystemEditor] Split StatEntry ushort into two separate bytes.

Zurphing:
Currently in KH2SystemEditor, The StatEntry category in the Item category is a ushort. It should be two separate bytes however, one for StatEntry, and the other APCost for abilities. This fixes an issue where abilities would have StatEntry's of 256, 512, 1024, allowing them to be more easily modified.

image
This image demonstrates the changes. Once More goes from having a StatEntry of 1024 (being read as 0x0400) to having an APCost of 4 (being read as 0x04)

### DIFF
--- a/OpenKh.Kh2/SystemData/Item.cs
+++ b/OpenKh.Kh2/SystemData/Item.cs
@@ -49,7 +49,8 @@ namespace OpenKh.Kh2.SystemData
             [Data] public byte Flag0 { get; set; }
             [Data] public byte Flag1 { get; set; }
             [Data] public Rank Rank { get; set; }
-            [Data] public ushort StatEntry { get; set; }
+            [Data] public byte StatEntry { get; set; }
+            [Data] public byte APCost { get; set; }
             [Data] public ushort Name { get; set; }
             [Data] public ushort Description { get; set; }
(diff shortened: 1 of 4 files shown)


TopazTK:
Yea, the "StatEntry" for abilities is supposed to be "Ability Type".
There are a whole bunch of elements here that should **not** be visible for the Ability category.

TopazTK:
Yea, the "StatEntry" for abilities is supposed to be "Ability Type" if I remember right.
In any case, there are a whole bunch of elements here that should **not** be visible for the Ability category. And in other categories, shit is mislabeled and not good to have. For instance, in some types the flags are one ushort, while in others they are one byte.
In some, flags are StatEntry and StatEntry are flags, and vice-versa.
This whole program needs rewriting. I am not approving these changes on these grounds. But if another contributor sees the need until this tool is reworked, they can approve in my place.

Zurphing:
Not something I thought to test before this, but it's worth mentioning that the PR would affect listpatching 03system/Item via Mods Manager, as APCost's can be edited more easily with this. Though, existing ItemList.ymls would have to be adjusted, which probably isn't ideal, especially for mods that edit a lot of items via this method.

PR #676, opened 2022-09-17
Panacea: Removing EOS override.

MainMemory:
I've been told it could cause legal trouble


TopazTK (approved):
I was just about to make a PR on this myself.

Yes, it can cause legal trouble because it is a DRM bypass. Looking at the code, it is actually one step further than that, it being a whole Epic Online Service Emulator.

It is very likely for OpenKH to face legal issues should the EOS Bypass is left in.
I am approving the changes, but I will let someone else merge should they want to have a say in this matter.

shananas:
Providing my opinion cause why not.
I personally am indifferent to the inclusion or lack therof of this feature without considering anything beyond can I run PC KH from OpenKH. It doesn't really save that much time.
But as people have said it provides a threat to openkh so considering that I would vote to remove it aswell. If it does in fact get removed I will go back through the mod manager code and edit everything appropriately. If not I will continue on the current edits I am making.
Considering everything better safe than sorry and I vote for removing this questionable feature.

Delta-473:
In my opinion it works like the Goldberg Steam Emulator which is a steam api emulator. https://github.com/GeospatialDaryl/Goldberemu This emulates Epic Online Services which in itself in a very limited capacity. Looking through the steam works dev docs they have there own DRM system that you need to enable specific apart from steam works online features. https://partner.steamgames.com/doc/features/drm If you check the Epic Online Services documentation on DRM: https://dev.epicgames.com/en-US/search?filter=Documentation&keyword=drm 
You don't find anything build into the sdk that's ready to go DRM.
The only check the game does is logging you in and calling 'EOS_Ecom_QueryOwnership' https://dev.epicgames.com/docs/api-ref/functions/eos-ecom-query-ownership
Didn't look into the game binary, just through the functions panacea implemented.

PR #678, opened 2022-09-17
Adding Unicode support to Panacea.

MainMemory:
This allows the mod folder to be placed in a directory containing Unicode characters.


kenjiuno (approved):
I checked successful build of Debug/Release, and some code checks with eye.

PR #680, opened 2022-09-18
Implement page stack and page back for ModsManager wizard

kenjiuno:
Implement page stack and page back for ModsManager wizard

https://user-images.githubusercontent.com/5955540/190887720-753197dc-7dc9-4c85-bb54-45ff283307ce.mp4

Fix #675

PR #681, opened 2022-09-18
Allowed PC to Build and Run or Run only

shananas:
If they have a valid shortcut to the KH launcher to speed up playing slightly. This will bring some merge conflicts with two other PRs from Kenjiuno. Its fine if those get accepted first so I can resolve the conflicts as I changed quite a few files for this


Vladabdf:
The aforementioned pull requests have been merged. #679 #680

PR #686, opened 2022-09-21
Add option to listpatch the cmd bar subfile in 03system.bin.

Zurphing:
Adds cmd as a possible type for listpatching via .yml files using the Mod Manager.

cmdList.yml files are setup exactly like TrsrList.yml files.
image

The ID of the command being edited as the "title", (this should match the ID of the command being edited) followed by the edits inside to change an actual command. In this example, the "Attack" command found in the command menu has its text changed to use the text for "Magic".

This should allow for compatibility between multiple different mods that change the cmd file.

### DIFF
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -569,8 +569,28 @@ namespace OpenKh.Patcher
                             }
                             Kh2.Battle.Plrp.Write(stream.SetPosition(0), plrpList);
                         break;
+                    case "cmd":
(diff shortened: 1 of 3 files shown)


Delta-473:
Can you add some documentation in https://github.com/OpenKH/OpenKh/tree/master/docs/tool/GUI.ModsManager in your PR.
Also if possible adding a test to https://github.com/OpenKH/OpenKh/blob/master/OpenKh.Tests/Patcher/PatcherTests.cs would also be nice you can use the existing ones as an example. :)

Delta-473 (approved):
LGTM

PR #692, opened 2022-09-24
Change FirstRun Variable and Quick Launch option

shananas:
Redid Alios' change to allow the wizard to run on an update. firstRunComplete is now wizardVersionNumber and if the new _wizardVersionNumber in MainViewModel.cs is greater than what is in mods-manager.yml the wizard is run again.

Moves Quick Launch option to its own menu that is now labeled so it is not a random dropdown box and also hide the Quick Launch Menu properly with no extra whitespace.


shananas:
cleaned up the code changes to a more compact better version. Ready to be merged now.

Vladabdf (approved):

PR #695, opened 2022-10-01
Recover auto release with GitHub Action

kenjiuno:
This will enable auto build and release of .NET projects, and also panacea, in zip file.

Create 2 release tags:
- `latest`
- `release2-${{github.run_number}}`

Sample from: https://github.com/kenjiuno/OpenKh/releases

2022-10-20_22h08_33

2022-10-20_22h08_24

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -2,9 +2,9 @@ name: .NET
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


MainMemory:
In order for the release to be complete, OpenKh.Panacea.dll has to be included in the output as well.

PR #700, opened 2022-10-24
Kh2MapStudio: EventScript experimental support (edit only)

kenjiuno:
2022-10-25_18h39_33


Delta-473 (approved):
Not tested myself but LGTM very exited for this!! :D

PR #710, opened 2022-11-11
Workaround for GitHub Actions `Error: Resource not accessible by integration`

kenjiuno:
2022-11-11_09h10_22

Thus apply `continue-on-error: true` to `marvinpinto/action-automatic-releases@latest`


kenjiuno:
ok merging this forcefully

PR #717, opened 2022-11-20
[Mods Manager] Add enmp as valid listpatch method

Zurphing:
Adds enmp as a possible type for listpatching via .yml using the Mod Manager.

The enmp file controls various stats for enemies, like elemental resistances, battle level, and health.
Enmp listpatches are set up like this: 
image

Documentation on how to listpatch the enmp subfile & tests have also been included.


Vladabdf (approved):
You love to see it.

PR #729, opened 2022-12-10
Made a menu item to show if openkh is configured for PC or Emulator

shananas:
To easier see if someone is asking for help what version Openkh mod manager GameEdition is. "PC Version" or "Emulator"

### DIFF
--- a/OpenKh.Patcher/PatcherProcessor.cs
+++ b/OpenKh.Patcher/PatcherProcessor.cs
@@ -106,7 +106,23 @@ namespace OpenKh.Patcher
                         if (assetFile.Required && !File.Exists(context.GetOriginalAssetPath(name)))
                             continue;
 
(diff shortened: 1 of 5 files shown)


shananas:
Copied from recent PR so its more visible.

Made local mod view for Zip mods able to show the desciption and preview image not only mods from github.

Also made ConfigurationService make a folder in "mods" for every supported game so installed mods will hide properly when swtiching to a game you havent yet installed a mod for.

shananas:
Commit msg copied again
Fixed mods manager only launching KH2 even if you dropdown launch game option was not KH2 on first startup. You used to have to select a different game then the one that was chosen if it wasn't kh2. I wasn't setting LaunchExecutable properly.

Also Fixed patching so it works with any of the 4 games tested with PC patches for each of the four games.
Made Restore only restore files for the currently selected game in the dropdown. It always restores Recom because it only has one .pkg and it is named differently.

Delta-473 (approved):
LGTM

PR #738, opened 2022-12-28
Addeed sklt to ListPatch with documentation and a test.

shananas:
Adds the ability to listpatch the sklt subfile in system.bin


shananas:
also sdded Atkp to Listpatch with documentation and tests.
have to append static info to the end of the atkp so it is the same size/info

shananas:
Added a null check on IsModBlocked to prevent Mods Manager from crashing when installing a github mod immediately after opening since the list it pulls from is currently empty

Vladabdf (approved):
Thanks for this addition

PR #749, opened 2023-01-04
Fix sklt listpatching.

shananas:
Game files use -1 when a character doesnt use a Bone 1 or 2 to hold a weapon so Bone 1 and 2 need to be shorts not ushorts


Vladabdf (approved):
Nice catch.

PR #751, opened 2023-01-06
Implement downloader and updater on client side

kenjiuno:
up2


kenjiuno:
ok this will be ready

TopazTK:
Maybe we should not reset the configuration upon update, and change the text in the window to say "Updating" instead of "Downloading Mod OpenKH"

kenjiuno:
I'll fix

kenjiuno:
This'll be better

TopazTK:
Everything looks good on the code side. Needs testing.

shananas (approved):
Works on my end too. I'll merge tomorrow if no one else does

PR #758, opened 2023-02-05
Mdlx importer

osdanova:
A working importer for the MDLX Editor!
Export an MDLX model to FBX, make a mesh replacement and import the FBX.

# Steps:
* Drag the MDLX to the tool to load it.
* File > Export as FBX.
* Work with the exported FBX. Example: Load it in Blender, delete the original mesh, add a new mesh and rig it to the skeleton, then export as FBX.
* Open the tool, load the original MDLX, and go to Import >Replace Mesh.
* You may need to tweak the limits, however the default values are usually the best.
* __If the import worked correctly the import window will close and the model will load. Otherwise an error message will show up and the window will remain open.__

Importer1
Importer2
Importer3

## Notes:
* This tool is designed to work by itself. Do not import FBXs that were not exported from this tool because chances are high of it not working.
* The textures have to be in the same folder as the model and have to be in PNG format.
* An FBX file can be dropped into the loaded MDLX to directly import using the default (Or current) settings as a shortcut, but errors won't be displayed.
* Blender exports FBX files with leaf bones by default. Since the tool only uses the original bones in the loaded MDLX, it ignores any extra bones and it __should__ work fine, but be careful with that in case any problems arise.
* Multiweighted rigs __should__ work, although they may be messy at times and the limits will have to be lowered.

## Skeleton:
The tool is ready to import skeletons, however the conversion from a Matrix4x4 to radians (Which is what KH2 uses) leads to wrong data, thus the resulting skeleton is a mess. This feature is disabled, if anyone is brave enough to try and fix it go to the class MdlxEditorImporter and change KEEP_ORIGINAL_SKELETON to false. The code to transform from Matrix to Radians is in the same file, in the function getSkeleton.


kenjiuno (approved):

kenjiuno:
Great to have mdlx importer!

PR #762, opened 2023-02-10
Complement CI build error from `dotnet publish`

kenjiuno:


Delta-473:
Wasn't there a breaking change with ImGui.NET starting from 1.8x?

kenjiuno:
> Wasn't there a breaking change with ImGui.NET starting from 1.8x?

Yes, you are right.
I know that previously there is a build problem by introducing layout editor with ImGui.NET 1.8+ pull request suggested by dependabot.

That problem occurred because error code from `dotnet build` or `dotnet publush` or such commands cannot be captured by CI build system.
This time I want to fix this problem at frist.
Next I'll fix incompatibility of ImGui.NET 1.8+.

To tell the truth, I have opened this pull request in wrong place `OpenKh/OpenKh`.
I had to open in my `kenjiuno/OpenKh`

Delta-473:
Hahaha happens, no problem :)

PR #768, opened 2023-02-19
MdlxEditor - Export Images

osdanova:
Images shown in the MDLX Editor can now be exported.

export

There's code (Written but hidden in the view) for adding/removing textures, however the textures are created from the data. Since the actual data remains the same, the output file will remain the same. This'll only work if the textures are changed in the actual data, for which there's no code as of now.


Delta-473:
can you pull/rebase the build fix in this pr? :)

Delta-473 (approved):
LGTM

Issue #545, opened 2022-01-28
[Feature Request] Linux Port


Pixal-Aqua:
Hey, I don't know if you can but could you try to make a port of OpenKH to Linux, because on Linux you can run openkh.modsmanager.exe but not put in the path to it because it's not a valid path.

Delta-473:
Hey, all the command line applications do run natively on Linux, the gui apps need to use wine because the gui framework used at the moment is WPF. To get them running natively on Linux it needs to be ported to MAUI which is quite a bit of work. The mod manager itself is even more complicated as it uses Windows specific functions to hook into memory and do the patching. However someone on the discord did made a basic version of it for Linux.

Pixal-Aqua:
Where would I this basic version of it like channel wise.

TauAkiou:
openkh.modmanager.exe will work, provided that you run it in WINE, with the .NET6 Desktop package installed and the `HKEY_CURRENT_USER/Software/Microsoft/Avalon.Graphics/DisableHWAcceleration` DWORD key set to 1.

The best shot right now seems to be running it through Lutris using the regular Epic Launcher, in which the mod manager WILL run.

The hangup appears to currently be Panacea, which is currently attempting to call MakeSureDirectoryPathExists from imghelp.dll. My guess is that WINE does not call this function properly - in fact, WINE doesn't seem to like it much when DBGHELP.dll is overridden.

Would it be possible to build a version of Panacea that injects through an alternative DLL, such as version.dll instead? That might have a better chance of guaranteeing Linux compatibility (and is also the injection path used by other modloaders, which seems to offer better compatability overall).

Kurumi78:
Everything seems to work from my testing on Linux with #788 being merged. Seems to be the issue could be closed unless there is another tool not working right.

TauAkiou:
I would argue that there is much more work that needs to be done. Panacea support is the first step, but there are a huge number of features and programs that don't work natively, which I believe is the target for proper Linux support.

Vladabdf:
I have heard all of the GUI programs work under WINE, including the Mod Manager now. Is that not the case? I already confirmed myself all CLI tools work properly eons ago, so I assume that is still the case.

PR #784, opened 2023-03-31
add dinput8.dll hooking support to Panacea

TauAkiou:
This PR adds the ability for Panacea to hook using 'DINPUT8.dll' as opposed to "DBGHELP.DLL". This method is used by LuaBackend, which is where I got the idea (and code hook information) to add support for this method of hooking.

Hooking this way plays more nicely with WINE, which would otherwise need a copy of 'imagehlp.dll' from a real windows installation to hook with the game due to passthrough conflicts with dbghelp.dll (many WINE dlls reduce code redundancy by forwarding to other DLLs for similar system calls). While this doesn't eliminate all issues with Linux support, it does help with issue #545 

Some testing has been carried out under Windows and Linux through WINE running a GoA seed.

Building in debug mode created a bunch of 'stack corruption' errors; I do not know if this occurs in standard releases of panacea outside of debug mode. Building in release suppresses them, and the DLL seemed to work regardless.

This PR will be left in Draft until some more testing (preferably by people other then me) has been carried out.

### DIFF
--- a/OpenKh.Research.Panacea/dllmain.cpp
+++ b/OpenKh.Research.Panacea/dllmain.cpp
@@ -10,6 +10,10 @@
 
 typedef BOOL(WINAPI* PFN_MiniDumpWriteDump)(HANDLE hProcess, DWORD ProcessId, HANDLE hFile, MINIDUMP_TYPE DumpType, PMINIDUMP_EXCEPTION_INFORMATION ExceptionParam, PMINIDUMP_USER_STREAM_INFORMATION UserStreamParam, PMINIDUMP_CALLBACK_INFORMATION CallbackParam);
 PFN_MiniDumpWriteDump MiniDumpWriteDumpPtr;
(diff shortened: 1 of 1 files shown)


TauAkiou:
i already know that I messed up the commits - if a maintainer wants to delete this PR so I can recreate it with a fixed commit chain (i messed up the DLL name and had to amend), that would be appreciated.

I did a little more testing - this change doesn't seem to affect hooking through "DBGHELP.DLL" at all, so it seems like the two hooking methods can coincide with no issue.

TopazTK:
DINPUT8.dll causes massive problems with the game's input code when tampered with, which leads to crashes. Lots and lots of crashes.
Notion to close this PR is now presented.

Delta-473:
I will close this, feel free to find an other dll file it can hook into, that doesn't result into issue. :)

TauAkiou:
i've written up one using "version.dll", which is a very popular hooking point for many, many projects. i'll test it out then post a new PR in the coming days.

PR #788, opened 2023-04-07
Panacea Hooking Support: VERSION.DLL

TauAkiou:
This PR, similar to #784, allows Panacea to hook to the game using "version.dll".

This hook is necessary to allow Panacea to play more nicely with WINE, improving Linux and Steam Deck support for the mod loader. 

An added side effect is that "version.dll" also hooks with "EOSSDK-Shipping-Win64.dll". This does not seem to interfere with operation as Panacea will automatically hang itself up, but whether or not the project is okay with this DLL hooking to an EGS dll is a question to ask.

Like before, this pull request will remain in draft until maintainers have had a chance to test or present questions.

### DIFF
--- a/OpenKh.Research.Panacea/dllmain.cpp
+++ b/OpenKh.Research.Panacea/dllmain.cpp
@@ -1,7 +1,13 @@
 #define WIN32_LEAN_AND_MEAN
+#define GetFileVersionInfoW GetFileVersionInfoW_
+#define GetFileVersionInfoSizeW GetFileVersionInfoSizeW_
+#define VerQueryValueW VerQueryValueW_
 #include <windows.h>
 #include <cstdio>
 #include <cassert>
+#undef GetFileVersionInfoW
+#undef GetFileVersionInfoSizeW
+#undef VerQueryValueW
 #define MiniDumpWriteDump MiniDumpWriteDump_
 #include <DbgHelp.h>
 #undef MiniDumpWriteDump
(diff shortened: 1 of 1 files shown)


Delta-473:
Doesn't break anything on Windows so LGTM

TauAkiou:
Opening up based on discussion here and on Discord.

To get this working on Linux, you will need to set the override for "version" to "Native". I am going to make a few more tests myself to make sure everything works.

There is also a WINE bug that manifests itself when running on Lutris where the console does not work properly. (it won't display anything related to the game, just WINE system messages). I'm not really sure how to get around that. It works fine when WINE is running through Heroic/Legendary, so i'm not really sure what's going on there.

Delta-473:
Idk if opening or asking wine devs would help solving the issue?

TauAkiou:
i'm not sure. my suspect is that this is related to this very old wine bug: https://bugs.winehq.org/show_bug.cgi?id=10941

personally, i'd recommend people use Heroic/Legendary to launch KH2 anyway. the console works properly when run through that.

it has the side effect of not being able to use Archipelago (if people were interested in that), but it's the recommended way to run EGS games on Linux anyway.

TauAkiou:
Panacea appears to be broken under Wine 8 but that can be easily worked around by moving back to a Wine 7 release. I am currently testing it using Heroic, which I have managed to get working reliably.

This may be something i'll have to investigate a bit deeper, since I'm unsure what changed in WINE that would be preventing the game from loading.

Delta-473:
Sad to hear that it's broken under wine atm, hopefully you're able to track it down easily, so that regression can be fixed. :) Thx for all the work you have put into it to improve Linux UX.

TauAkiou:
to clarify, Panacea is only broken under Wine 8. you can easily move back to wine 7 in Heroic and it will work perfectly fine again.

this particular regression is annoying, as Wine 8 fixes a bunch of problems with the modloader (data extraction works reliably, no longer requiring hardware acceleration to be disabled, building works properly, etc). it would be nice to have an all-in-one solution that Just Works(tm).

for now, switching between wine-ge-8 and wine-ge-7 seems to be the best solution. 

An additional positive: I figured out a workaround for getting Archipelago working with KH2 on Heroic. I did a full sync run last night with only one crash in STT. (and some controller issues, which I was able to work around). Still uses the "VERSION.DLL" version of Panacea, which works without any proprietary windows DLLs.

Delta-473 (approved):
LGTM

Delta-473:
As other issues are wine related this can be merged.

Kurumi78:
I know this is already merged, however I would like to comment that Panacea seems to work fine on Wine 8, (I'm using Wine-GE 8-10). The thing thats breaking is LuaBackend which seems to cause a hang on Wine 8.

PR #789, opened 2023-04-07
Bdx: Fix invalid processing of decimal parser/ToString in French culture setting

### LINKED ISSUE #785: [BUG] - Mod Manager bdscript error
**OpenKH Build**
Running from current master so OpenKH Build 453

**OpenKH Program**
OpenKH Mod Manger

**Describe the Bug**
bdscript method to make an AI change from a bdscript gives an error if computers language is not in English. Not sure how many languages are affected but switched my pc to French and was able to reproduce the issue on my first Boss/Enemy seed.

**To Reproduce**
Change the language of windows to French (maybe other languages too)
Use the 3.0 v3 seed generator for KH2 Rando to create a seed with Boss/enemy Rando (i used one to one on both. unsure if some seed might work and some don't first try I got the error.)
install seed into mod manager and do any kind of build and receive Input string was not in the correct format error.

**Expected behavior**
Seed builds and AI for changed bosses/enemies works normally as it does if you PC language is English

**Screenshots**
Other mods are irrelevant reproduced in debug in visual studio with only the rando seed installed but quickly got lost in how bdscripts works
image
Error happens starting on line 129 in BdxEncoder.cs
image

**Additional context**
@kenjiuno Since you made the ai stuff I think @snelson3 since you implemented it into mod manager 
Just after posting this switched my language back to English and seed built perfectly fine.

kenjiuno:
Fix #785


shananas (approved):
Tested with both French and Norwegian as thats was the two languages people had posted this problem with works fine now.

PR #833, opened 2023-07-26
Fix a swapping problem of R and B channels on OpenKh.Engine.MonoGame

### LINKED ISSUE #825: [BUG] - Tool(s) have inverted colors
**OpenKH Build**
520

**OpenKH Program**
OpenKH.Tools.LayoutEditor
(possibly more) 

**Describe the Bug**
Colors on opened files with images are inverted (likely only imd & imz)

**To Reproduce**
1. Open program
2. Open file in program
3. See issue

**Expected behavior**
Colors should not be inverted. 

**Additional context**
This issue happens every so often. Are there code regressions happening under the hood that are not being documented? This must be the 10th+ time this has happened. 
For context, I only had time to test a Re:CoM layd + imz combination. I did not test KH2 or more than two additional files. Historically, when this issue happens, it occurs in all tools since they share common libraries.

kenjiuno:
Before:

2023-07-26_20h42_54

After:

2023-07-26_20h42_29

---

`SurfaceFormat` that RenderTarget2D uses in default is `Color`.

https://github.com/MonoGame/MonoGame/blob/b21463b419e55b4c898030fc22bee77dabb11210/MonoGame.Framework/Graphics/RenderTarget2D.cs#L65-L66

https://docs.monogame.net/api/Microsoft.Xna.Framework.Graphics.SurfaceFormat.html

> Color	
> Unsigned 32-bit ARGB pixel format for store 8 bits per channel.

I'm not sure whether this means that the byte order is `AA RR GG BB` or not.

Fix #825

### DIFF
--- a/OpenKh.Engine.MonoGame/MonoSpriteDrawing.cs
+++ b/OpenKh.Engine.MonoGame/MonoSpriteDrawing.cs
@@ -180,7 +180,7 @@ namespace OpenKh.Engine.MonoGame
         {
             var size = image.Size;
             var texture = new RenderTarget2D(_graphicsDevice, size.Width, size.Height);
-            texture.SetData(image.AsBgra8888());
+            texture.SetData(image.AsRgba8888());
 
             return new CSpriteTexture(texture);
         }


Delta-473 (approved):
lgtm

PR #843, opened 2023-08-06
Implement Texture and Mesh Order Switching

Joe-Degler:
Allows the user to change the order of meshes and textures (but not texture animations) of an mdlx file. Keeps Texture & Texture Animation associations in order, so that the .tim file stays valid (as far as my testing could tell).

Foreword: It's been ages since I wrote C#, and I never have written much WPF to begin with.

image
image

For Textures, switching order is as simple as changing the order of the `TextureModel` and adjusting a couple indexes.

For Meshes, this is more complicated. For some reason I don't quite understand, messing with the `Mesh` order causes the file to become corrupt when saved as MDLX - the file cannot be opened in other viewers or used in the game (the game crashes). 

Luckily, when we export the modified mdlx as fbx and reimport the same fbx, the model is now exportable as a valid mdlx. So internally, we do exactly that when switching the order. We do have to take a middle step and store it in an temporary file, because apparently during the export process the `scene` still gets modified in a way necessary to make it not corrupted. 

In the future if we figure out why the mdlx is corrupt when the mesh order is changed, we can get rid of this ugly process.

I'm a bit strapped for time, so I just reused the settings menu of the Import Window right now, but if anybody has better ideas how to do this, feel free to suggest or implement.

### DIFF
--- a/OpenKh.Tools.Kh2MdlxEditor/Views/Main_Window.xaml.cs
+++ b/OpenKh.Tools.Kh2MdlxEditor/Views/Main_Window.xaml.cs
@@ -120,7 +120,8 @@ namespace OpenKh.Tools.Kh2MdlxEditor.Views
                     contentFrame.Content = new ModelFile_Control(mainWiewModel.ModelFile);
                     break;
                 case Bar.EntryType.ModelTexture:
-                    contentFrame.Content = new TextureFile_Control(mainWiewModel.TextureFile);
+                    // TODO the second parameter should be the texture file, but it's not available here. 
+                    contentFrame.Content = new TextureFile_Control(mainWiewModel.TextureFile, null);
                     break;
                 case Bar.EntryType.ModelCollision:
                     contentFrame.Content = new Collision_Control(mainWiewModel.CollisionFile);
--- a/OpenKh.Tools.Kh2MdlxEditor/Views/Model_Control.xaml
+++ b/OpenKh.Tools.Kh2MdlxEditor/Views/Model_Control.xaml
@@ -39,12 +39,13 @@
         <Grid Grid.Column="2" Background="#202020">
             <Grid.RowDefinitions>
                 <RowDefinition Height="*" />
+                <RowDefinition Height="Auto" />
                 <RowDefinition Height="2" />
                 <RowDefinition Height="*" />
             </Grid.RowDefinitions>
             
             <!-- Meshes -->
-            <ListView Grid.Row="0" ItemsSource="{Binding Groups}" Background="#1d1d1d" Foreground="White">
+            <ListView Name="LV_Meshes" Grid.Row="0" ItemsSource="{Binding Groups}" Background="#1d1d1d" Foreground="White">
                 <ListView.Resources>
                     <Style TargetType="GridViewColumnHeader">
                         <Setter Property="Visibility" Value="Collapsed" />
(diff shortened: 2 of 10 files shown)


OpenKh.Tools.Kh2MdlxEditor/Views/Main_Window.xaml.cs
```diff
@@ -120,7 +120,8 @@ public void openBarEntry(Bar.Entry barEntry)
                     contentFrame.Content = new ModelFile_Control(mainWiewModel.ModelFile);
                     break;
                 case Bar.EntryType.ModelTexture:
-                    contentFrame.Content = new TextureFile_Control(mainWiewModel.TextureFile);
+                    // TODO the second parameter should be the texture file, but it's not available here. 
+                    contentFrame.Content = new TextureFile_Control(mainWiewModel.TextureFile, null);
```
Joe-Degler, line 124: This file might be deprecated entirely? Everything seems to use Main2.

OpenKh.Tools.Kh2MdlxEditor/Views/Model_Control.xaml
```diff
@@ -84,9 +85,20 @@
...
+            <DockPanel Grid.Row="1">
+                <!-- Buttons to reorder the meshes -->
+                <Button DockPanel.Dock="Right" Margin="5" HorizontalAlignment="Right" Width="40" Height="25" Click="Mesh_MoveDown"
+                        ToolTip="Move the selected mesh up. This uses an fbx export and import behind the scenes, make sure the settings in `File > Import > Replace Mesh...` are correct." >
+                    \/
+                </Button>
+                <Button DockPanel.Dock="Right" Margin="5" HorizontalAlignment="Right" Width="40" Height="25" Click="Mesh_MoveUp" 
+                        ToolTip="Move the selected mesh down. This uses an fbx export and import behind the scenes, make sure the settings in `File > Import > Replace Mesh...` are correct." >
+                    /\
+                </Button>
```
Joe-Degler, line 97: Would be cooler if it was drag&drop instead. The switcher is offset based, so it certainly would be possible.

osdanova:
I'm seeing a couple of things I don't like very much (Using main as main's constructor and having to create an fbx file and read it to reorder the meshes). Let's see if we can fix the reordering first and then it'll be much cleaner :)

osdanova:
I've been testing this PR myself and it works. Due to how the PS2/KH2 works, you can't simply reorder the meshes, since they are loaded contiguously in memory, thus reordering requires remaking the whole model from scratch. I've tried doing so myself and I haven't been successful, I have no idea of what exactly is not working in the building process. The next best thing is creating an fbx and then reading it, which is what this PR does.
Since the version in the PR works and we have nothing better I think this is alright.

NOTE: This process creates a temp file for the FBX, I'm not sure what happens with it but I think we should be deleting it after use right? That's the only change I think we need in this PR.

shananas:
If this does make a temp FBX and never delete it then yeah it definitely should be tracked and deleted afterwards. I'd Assume its in a temp folder in appdata (new one each time usually).

Joe-Degler:
While I was using my tool for other models (in this case Roxas), I came into trouble with bitmap imaging screwing up the colors still. I figure this approach isn't good enough for the textures as well, and might also require fbx exporting and reimporting.

I've had more success using https://github.com/BobbyAnguelov/FbxFormatConverter to convert a blender .fbx to ascii (and then back to binary), going to `Object connections` and reordering the meshes manually.

Texture order in MDLXViewer is determined by which order the meshes appear in `Object connections` - And I don't know what exactly causes blender to determine in what order they're exported (It's definitely not name, it's some internal ID). 

With that method I've had 0 problems making things work.

Just some food for thought, I'll revisit this if it's still necessary (and not covered by the other tool @osdanova posted about in Discord).

PR #844, opened 2023-08-07
Fix png loader of OpenKh.Imaging.PngImage

kenjiuno:
Fix png loader of OpenKh.Imaging.PngImage: crash when tRNS (transparency or alpha) chunk is missing on indexed image.

`OpenKh.Tools.ImageViewer` is one of affected tools.

2023-08-07_09h47_43


Vladabdf (approved):
Thank you for the fix!

PR #849, opened 2023-08-13
Updated readme imagery, opinion wanted

Vladabdf:
Updated the runtime image, made a slightly more focused and clearly depicted architecture breakdown. Also improved the non-Windows GUI note and moved it under the runtime download image.

image


shananas:
Changes look good to me. Do kind of question how much hope there is for openkh game engine. Seems superseded by the PC port but I mean it does still exist in openkh so someone could pick it up one day.

Vladabdf:
The game engine was something always intended to be far beyond the initial scope of the project. It's a massive undertaking, but the outline is there which does indeed leave it available to be picked up later if need be. It's ambitious but still outside our reach for the time being. Nevertheless, the point here was to just make something slightly more coherent to display in the Readme.

shananas (approved):
Looks good willing to merge but will wait a little longer for possibly more opinions.
Just noticed the .NET installation image is a different background color. Doesn't matter and I think its a nicer color on the eyes anyways just mentioning it.

PR #858, opened 2023-08-31
Mod Manager Lua engine integration and Presets

shananas:
Adds Presets and Lua Engine integration.
When saving a new preset you will need to restart mod manager for the preset to show under Load Presets. If you save a preset with the same name as one that already exists it will override it and a restart will not be necessary. If you try to include characters in the name that windows does not let be in a filename it will automatically be replaced by +.

Adds an install Lua Engine option into the setup wizard if you chose PC version. Can either download and configure Lua Engine to automatically load scripts installed in Mod Manager. Or if you have it installed already can locate where Lua Engine is installed and use it from there and configure it to load scripts from Mod Manager. 
Can also install a Lua Script directly into mod manager as an easy way to enable or disable them.
image
When you install a script it will generate a mod.yml based on MetaData in the Lua or using defaults. The script will be copied to be in `mod/"chosen game"/scripts` which the Lua Engine configuration step in the setup wizard will set it so Lua Engine can load scripts from there. If you have Lua Engine installed either through Mod Manager or by selecting where you installed it and use Build and Run or Run only it will load the game and then load start Lua Engine.

Taken from osdanovas PR to my fork.
Changes the presets to its own menu. Backend remains mostly the same.
266825917-7eaac0ef-ba31-428d-9c83-8c068b5a1a1f


osdanova:
I haven't messed around with presets but I don't think the basic structure for them is ready.

1. Reloading the app is unnecessary, simply assign the MenuItem's ItemSource to an ObservableList<MenuItem> in the ViewModel. Then, when the app loads, initialize and fill the list and whenever a new preset is saved add to that list.

Presets

2. I do not think not being able to remove presets is acceptable.

TopazTK (changes requested):
Please add a way to remove presets.

shananas:
Delete confirmation Box. everything else thanks to osdanova
image

TopazTK (approved):

TopazTK:
LGTM!

shananas:
Back to draft until Lua Engine and common Lua scripts work nicely together

TopazTK:
I am happy with the state of the current LuaEngine, but give time for 1-2 more releases.

PR #859, opened 2023-09-01
Auto Update toggle added to info menu in Mod Manager.

shananas:
When checked/true Mod Manager will update all installed mods on startup if there are any updates found.

image


shananas:
> Looks great. However, could you make it say "Auto update mods" just to be a little more clear about what's being updated?

Done

Vladabdf (approved):
Much appreciated

PR #861, opened 2023-09-03
Panacea: Fixing Gummi crash, implementing BASS_VGMStream for KH2 titl…

MainMemory:
…e/results/gummi music.

I did not modify the build process at all. All the DLL files in the OpenKh.Research.Panacea/Dependencies folder need to be copied to the build, as well as being copied into the game's folder alongside the Panacea DLL (DBGHELP.DLL).

PR #862, opened 2023-09-04
Mod Manager update for new panacea 

shananas:
with error handling to prevent a partial panacea install causing the game to crash.

If Panacea is missing files it will remove Panacea and any partially installed dlls and you will be unable to use the mod loader. Otherwise install works normally. Checksum is still only calculated for Panacea not these dlls.
image

this adds an additional 13 dlls that Panacea needs to the game install folder. The remove command will remove all 13 + Panacea dlls


shananas:
Merge this before #861 then merge #861 right after. This way if someone downloads the build from this merge it has error handling. If done the other way and user installs from the Panacea merged release the game will crash.

Vladabdf (approved):
Awesome! Thank you!

PR #867, opened 2023-09-08
Commandline build and patch operations

tryashtar:
Thanks for the great work on this project. As part of some work I was doing, I was interested in automating the process of swapping out selected mods and building them, without having to manually click through the mods manager.

I couldn't find a way to do it, so I added it myself! :) I hope it's satisfactory; do let me know if there is anything important I missed when trying to contribute for the first time.


Vladabdf:
I completely missed this and your message in the [Discord server](discord.openkh.dev), but isn't this effectively what `OpenKh.Command.IdxImg` does?

shananas:
In terms of adjusting the setup wizard to understand alternate panacea destination location I believe the code should be this.
```
if (!File.Exists(PanaceaDestinationLocation) && !File.Exists(AlternatePanaceaDestinationLocation))
{
    // DBGHELP.dll is not installed
    PanaceaInstalled = false;
    return false;
}```
```
return IsEqual(
    CalculateChecksum(PanaceaSourceLocation),
    CalculateChecksum(PanaceaDestinationLocation)) || IsEqual(CalculateChecksum(PanaceaSourceLocation),
    CalculateChecksum(AlternatePanaceaDestinationLocation));
    ```

tryashtar:
Looking at `IdxImg`, it may well be! I'll have to check and see if I can replicate my needs with that one.

And regarding the alternate Panacea code, I agree that's the more natural way, but it no longer sets `PanaceaInstalled` to match the final `return` value that way. I'm not sure why that side effect is part of the function, so I tried to preserve that behavior.

shananas:
Looking back at it I did miss something id still keep my top code and skip the bottom else so
```
 if (!File.Exists(PanaceaDestinationLocation) && !File.Exists(AlternatePanaceaDestinationLocation))
 {
     // DBGHELP.dll is not installed
     PanaceaInstalled = false;
     return false;
 }
 some code
  if (File.Exists(PanaceaDestinationLocation))
                {
                    return IsEqual(
                        CalculateChecksum(PanaceaSourceLocation),
                        CalculateChecksum(PanaceaDestinationLocation));
                }
                else if (File.Exists(AlternatePanaceaDestinationLocation))
                {
                    return IsEqual(
                        CalculateChecksum(PanaceaSourceLocation),
                        CalculateChecksum(AlternatePanaceaDestinationLocation));
                }```

tryashtar:
I believe that version of the code would crash if panacea is in neither location, since `CalculateChecksum` expects the file to exist

tryashtar:
I looked a bit closer at `IdxImg`. I was able to find it has the game file extraction, which is helpful... but I don't see anything that replicates `PatcherProcessor` (reading the mod metadata and switching on all the different asset types).

If it helps, the code from the mod manager that I want to access from the command line is basically this:
* https://github.com/OpenKH/OpenKh/blob/895b7624f668144120dbedba281d3cf4084fe24b/OpenKh.Tools.ModsManager/Services/ModsService.cs#L343-L369

This one may be useful too. Am I correct in my understanding that the above code is for when Panacea is being used, and the below code is for when it's not?
* https://github.com/OpenKH/OpenKh/blob/895b7624f668144120dbedba281d3cf4084fe24b/OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs#L632-L792

tryashtar:
There, I added a `patch` command as well that replicates that section. This seems to cover everything I need :)

tryashtar:
Hi, any thoughts on this? I know the panacea part was superceded by #918, but I still really am in need of a commandline patcher. It doesn't have to be this, but it's ready and working. Any solution would be much appreciated :)

shananas:
I don't feel qualified with the knowledge of what the patcher already does vs this PR but regardless first the merge conflicts would need to be resolved and as major changes have been made I belive it just amounts to scrapping any mod manager changes I believe specifically the setupwizard changes as I have satisfied the Linux problem of mod loader.
I would rather another org who is more knowledgeable make the call whether to accept the patcher portion of the PR.
If my understanding is correct this is just a way to after mods are installed into mod manager to build them for panacea through a command line interface

tryashtar:
I can throw out the superceded changes if that would make merging easier. Probably needs to be done either way

shananas:
Also to show my lack of knowledge does the patcher not have the ability to patch them game.
Basically for it to be fully compatible/work kind of how mod manager works there is a difference in how nods are built for panacea and how they are built for patching. Can this build it both ways or is this introducing a way to build for panacea and one already exists for patching?

tryashtar:
The `build` step builds the mods. If you're using panacea, you're done here. If not, then you use the `patch` step to finish. This PR provides both

Vladabdf:
> I looked a bit closer at `IdxImg`. I was able to find it has the game file extraction, which is helpful... but I don't see anything that replicates `PatcherProcessor` (reading the mod metadata and switching on all the different asset types).

IdxImg has an `inject` function. Is that not what you're effectively trying to accomplish here?

tryashtar:
You mean [this][https://github.com/OpenKH/OpenKh/blob/16dd0ff512d5247390d1fcf1f340d310466a2a80/OpenKh.Command.IdxImg/KingdomHearts2.cs#L142]? To my understanding that's only for PS2 ISOs; I don't see an equivalent in `EpicGamesAssets`.

shananas:
> > I looked a bit closer at `IdxImg`. I was able to find it has the game file extraction, which is helpful... but I don't see anything that replicates `PatcherProcessor` (reading the mod metadata and switching on all the different asset types).
> 
> IdxImg has an `inject` function. Is that not what you're effectively trying to accomplish here?

looking at idximg `inject` looks to be specifically for an iso/ps2 version which wont work as this if for the PC version.

comparing this to idximg's EpicGamesAssets.cs this introduces a way to build the mods in mod manager through command line so linux users or even those who dont wish to open mod manager can use command line. then for the patching this from what i can tell will do all at once while looking at idximg it would be one pkg at a time.

tryashtar:
I've moved the code to IdxImg. I hope this is satisfactory.

shananas:
It seems there is still merge conflicts to be resolved idk if your branch is out of date from main causing a different hash? for KH2Object editor or what's exactly it. Also since this PR was opened Mod Manager now has support for KH3D idk if it'd be as simple as allowing that to be the chosen game but support for that should be added if possible.
I would still like to ask someone more knowledgeable with IdxImg to approve rather than me but this is definitely different functionality that previous. @Vladabdf possibly since they've been commenting on this already.

shananas (approved):
Ill leave this for a little longer but will approve for now and merge after a few more days if no one else comments.

tryashtar:
Much love, thank you! And feel free to change how this works at any time, I'm not tied down to any backward compatibility concerns

PR #868, opened 2023-09-09
UV mapping fix

osdanova:
In WPF texture UVs (Which go from 0 to 1) take 1 as the highest value in the mapping list, meaning meshes that don't have UVs that go from 0 to 1 would render incorrectly (Thanks WPF)
Whenever the model is rendered in WPF a couple of dummy vertices are added with UVs 0,0 and 1,1 so the mapping scales correctly, which fixes the issue. Now all models should be rendered accurately.

image

image


Oathseeker:
Test0
Test1
Test2


I can confirm that the UV issue is fixed, the model preview shows correctly in both the mdlx editor and obj editor, where as before the preview was broken as shown in a previous version of the tool.

Oathseeker (approved):
UV mapping is working correctly, verified it with older release vs newer release.
Test0
Test1
Test2

shananas (approved):

PR #872, opened 2023-09-11
Add a way to add Sprites in the Layout Editor [SEQUENCE]

TopazTK:
Because I need it, too.


TopazTK:
Added a way to fix UV animations whilst on it.

Vladabdf (approved):
My precious layout editor has reached a new peak. Godspeed to ya, and thanks for the contribution

PR #876, opened 2023-09-12
Gives user the option to overwrite old mod install.

shananas:
In case it partially installed or has been edited and they want to install fresh. They do not have to remove the mod then install can just reinstall. Very helpful if a partial mod install happened because sometimes the mod didn't show up in mod manager so you had to go into the `mods` folder and delete it.

Zip
image
GitHub
image
No Click
image

Yes proceeds like normal after deleting the old mod install.


TopazTK (approved):
Based.

PR #878, opened 2023-09-14
Try to fix #877

### LINKED ISSUE #877: [BUG] - Newly added sprites in Tools.LayoutEditor crash the program when deleted
**OpenKH Build**
643

**OpenKH Program**
OpenKH.Tools.LayoutEditor 

**Describe the Bug**
Adding sprites in the sprite editor function as intended, however deleting said new sprites causes the program to crash with no error. 

**To Reproduce**
1. Open program
2. Open sprite editor sub-window 
3. Click the arbitrarily unlabeled `+` button near the top of the window to add a new sprite
4. Select the new sprite index
5. Click the arbitrarily unlabeled `-` button near the top of the window to delete the currently selected (new) sprite
6. Cry after program has crashed and progress potentially lost

**Expected behavior**
The sprite should be deleted and the selected sprite index should reverse by 1 or move forward by 1 if none behind. 

**Additional context**
I cannot currently test, however I believe I know the cause of the crash. When deleting an existing sprite, every sprite with a higher index is moved down by 1 place and the program then reads the information about that next sprite. When the newly added sprite is deleted, there is nothing to decrease the index # by, and therefore no information to read.

**Potential fix**
Upon a sprite being selected, the program should check in the background to see if there is a sprite both ahead and behind in the index. If one is located, upon deletion the selected index should move that direction to avoid the crash. If none is detected, do not allow the sprite to be deleted.

TopazTK:
Fixes #877

### DIFF
--- a/OpenKh.Tools.LayoutEditor/Dialogs/SpriteEditDialog.cs
+++ b/OpenKh.Tools.LayoutEditor/Dialogs/SpriteEditDialog.cs
@@ -59,11 +59,14 @@ namespace OpenKh.Tools.LayoutEditor.Dialogs
                 DrawCropAtlasTexture();
             }
 
-            if (ImGui.Button("+"))
+            if (ImGui.Button("Add Sprite"))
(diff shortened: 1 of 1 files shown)


Vladabdf (approved):
Perfect. Thank you very much

PR #880, opened 2023-09-15
Mdlx Builder - Fix import if no vertex colors

osdanova:
When implementing vertex coloring, it was enforced accidentaly. If the model didn't have vertex coloring the app would throw an error. Changed it so that vertex color is only applied if all the vertices have vertex coloring.


Delta-473 (approved):
Didn't test but LGTM

PR #882, opened 2023-09-17
Panacea Dependency fix

shananas:
Update dotnet.yml with suggested xcopy


shananas:
it works just checked the release from the branch

Vladabdf (approved):

PR #886, opened 2023-09-18
Disallow Mod Manager from working with absolute directories.

### LINKED ISSUE #857: [BUG] - OpenKH Mod Manager can copy files into unintended places
**OpenKH Build**
Latest release, release2-599

**OpenKH Program**
OpenKh.Tools.ModsManager.exe

**Describe the Bug**
The Open KH mod manager is able to arbitrarily copy and replace files anywhere on your computer using the mod.yml file. Obviously without admin permissions some locations are off limit, but if you run mod manager as admin those work too. Variables like %appdata% and %UserProfile% also seem to kinda work. The program crashes which means this was either unexpected behavior or someone tried to avoid this, but it still manages to copy all the files you specify.

**To Reproduce**
1. Install a questionable mod from URL or a user supplied ZIP
2. Enable the mod
3. Build and patch

**Expected behavior**
The program should check that the folder it's placing files in is a relevant folder and throw an error, or at least crash BEFORE copying files.

**Example**
https://youtu.be/wUHw2WjIWPM

**Additional context**
Similar methods of placing files and folders in arbitrary places can happen when using / or ../ when downloading from a URL. For example /KH2FM-Mods-Num/GoA-ROM-Edition successfully creates a folder and clones everything to the root of whatever drive mod manager is running on, and ../../KH2FM-Mods-Num/GoA-ROM-Edition creates a folder in the same directory as mod manager but luckily fails to download and gives a file not found error.

TopazTK:
This is to ensure that THIS doesn't happen: https://www.youtube.com/watch?v=wUHw2WjIWPM
Fixes #857. Thanks @adamboy7.


Vladabdf (approved):
Approved, but since this is such a major increase in security, I am requesting at least two other reviews before it get merged.

TopazTK:
> Approved, but since this is such a major increase in security, I am requesting at least two other reviews before it get merged.

Major İncrease in Security with 3 lines of code :^)

Vladabdf:
It's still a massive oversight that went overlooked for awhile. I'm glad it got fixed

shananas (approved):
LGTM

osdanova (approved):
LGTM

tryashtar:
This doesn't seem to protect against paths like `..\..\..\..\..\..\..\..\blah.txt`

MainMemory:
Safest bet is probably to use `Path.GetFullPath` and check that the result starts with the path to the mod folder.

Vladabdf:
Correct, absolute paths have been disabled. At this point since all patching is relative, the absolute best we can do is tell people to not use mods they do not trust and to check the "Files to patch" info drop down on the right side of the mod manager. If you've installed a mod that includes something ridiculous like `../../../../../../hack.exe`, then there are two points of concern:

1. They cannot possibly guess where your root directory or crucial info like the Windows installation directory are relative to where you are running OpenKH, and
2. It is ultimately up to the end user to be their final barrier of protection against attack from someone malicious, and if they install something without knowing what it will do, a computer is perhaps not a toy the user should be playing with. 

We can only do so much, and because of how the mod manager works by design, I cannot imagine there is a way to fool-proof this further, unless it were possible to outright ban the use of `../` relative pathing.

shananas:
I mean I see no reason to allow `../` all mods should begin and end in `mod/selectedgame`

PR #889, opened 2023-09-19
Vertex normals, Vif code update and MDLX importer update

osdanova:
Behold! The last MDLX importer update! (I hope)

* The VIF code has been updated and it now should reflect the data more accurately.  This seems to have fixed an issue with the previous importer release.
* The importer now supports vertex normals.
* Vertex colors and normals can be selected to be applied per mesh.
* New buttons have been added to open a file by selecting it instead of drag & dropping and to save over the opened MDLX.
* If I have to touch this again I'm gonna jump out the window.

image

When importing, the user can either select the file or drag and drop it. Once the file is loaded the list of meshes will appear. For those meshes that have colors or normals, their checkbox will be able to be checked to import the colors or normals. Once everything's set, clicking in Import will finish the process.

Exmaple of a model with normals and colors applied:

image

### DIFF
--- a/OpenKh.AssimpUtils/Kh2MdlxAssimp.cs
+++ b/OpenKh.AssimpUtils/Kh2MdlxAssimp.cs
@@ -234,6 +234,12 @@ namespace OpenKh.AssimpUtils
                         iMesh.VertexColorChannels[0].Add(new Assimp.Color4D(R, G, B, A));
                     }
 
+                    // NORMALS
+                    if (vertex.Normal != null)
+                    {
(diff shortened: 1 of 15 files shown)


Oathseeker:
Heck yeah, thanks so much for your work on this, I will try to test it later tonight.

Oathseeker (approved):
Working fine for me, also the issue where the model would crash or bug out if it didnt have vertex colors is fixed.

TestScreenshot1

Oathseeker (approved):
Why it still say 1 review needed I approved :( ?

Vladabdf (approved):

PR #891, opened 2023-09-20
add back BinarySharp and Hypervisor

shananas:
Pulls files from before accidental removal.


Vladabdf (approved):
Thanks for undoing the minor implosion

PR #899, opened 2023-09-29
Presets and Lua Script install.

shananas:
Took everything from Lua Engine PR into its own. I believe this is better to have in main and once Lua Engine is ready it will be merged by itself. will copy relevant info from mention PR below.

Add Presets and Lua Script install.
~~When saving a new preset you will need to restart mod manager for the preset to show under Load Presets.~~ If you save a preset with the same name as one that already exists it will override the previous preset. If you try to include characters in the name that windows does not let be in a filename it will automatically be replaced by +.

Taken from osdanovas PR to my fork.
Changes the presets to its own menu. Backend remains mostly the same.
266825917-7eaac0ef-ba31-428d-9c83-8c068b5a1a1f

Can also install a Lua Script directly into mod manager as an easy way to enable or disable them.
image
When you install a script it will generate a mod.yml based on MetaData in the Lua or using defaults. The script will be copied to be in `mod/"chosen game"/scripts`. Without Lua Engine this is mostly useful for KH2Rando. But anyone with Lua Backend installed and configured to load scripts from `mod/"chosen game"/scripts` as the KH2Rando seed gen configuration does for the user can use this.

### DIFF
--- a/OpenKh.Research.Panacea/OpenKh.Research.Panacea.vcxproj
+++ b/OpenKh.Research.Panacea/OpenKh.Research.Panacea.vcxproj
@@ -66,6 +66,9 @@
       <EnableUAC>false</EnableUAC>
       <AdditionalDependencies>Shlwapi.lib;Dependencies\bass.lib;Dependencies\bass_vgmstream.lib;%(AdditionalDependencies)</AdditionalDependencies>
     </Link>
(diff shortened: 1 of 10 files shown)


Vladabdf (approved):
Excellent work

PR #902, opened 2023-10-01
Various changes.

shananas:
Delete the downloaded zip for the update from temp AppData after extracting it. 
Add more lines to the .bat. Sometimes window?.close() fails to close Mod Manager because window is null so the bat will try also before doing anything. If MM is alreaddy closed the bat just continues to execute as normal. After starting the updated MM deletes the extracted openkh download then itself. 
Preset window changed to ShowDialog() so you cannot mess with the main window while the preset window is open. 
Moved Check for update to the settings tab.
image


Vladabdf (approved):
All sensible changes. Thanks

PR #905, opened 2023-10-06
Fixed Mod Updates being hidden and added ToolTips for 4 buttons on Mod Manager main window

shananas:
ToolTips for the MoveUp, MoveDown, AddModCommand, and RemoveModCommand.
Changed instances of ReloadModsList() to only do whats needed when possible. If not possible FetchUpdates after ReloadModsList() so the updates do not get hidden until a Mod Manager restart.


shananas:
GUI for panacea settings. Also do not close menu when checking options with a checkbox. 
I dont GIF so heres images instead.

Default
image

image

Debug Log wont show if Console isnt enabled so it being checkable requires Console enabled and when unchecking console Debug Log also becomes unchecked
image

TopazTK (approved):
Based!

PR #918, opened 2023-10-29
Proper version.dll support in mod manager and preset save fix.

shananas:
This does take a portion of setup wizard code from #867 just added a little more detection for an edge case if files with both names are found somehow.

image

New commit adding check for mods update command so the user doesnt need to restart MM to check mods for updates if they are not showing.

image


shananas:
Didn't make this to circumvent tryashtars PR. This is full MM support with an install button while that was just preventing MM from improperly setting Panacea to false if you ever re-ran the setup wizard on Linux. Also as I'm unsure of their patcher changes I cannot approve/merge the whole PR just to have the checksum panacea check.
This does make the setup wizard code edit of their PR unneeded.

Vladabdf (approved):

PR #929, opened 2023-11-12
KhModels and data structures for KH1 and DDD

osdanova:
KhModels is a tool that allows to open all kinds of models from the series and export them as fbx for whatever purpose is needed. 

Currently it can open and export the following formats:

KH1:
- MDLS (No HP /Multiweight)
- WPN

KH2:
- MDLX
- MAP

BBS:
- PMO
- ARC

DDD:
- PMO (No textures)
- PMP (No textures)

image

How to use:
- Drag and drop one of the files listed above
- Right click to rotate, and mouse wheel to move
- Export > FBX to export the model
- Mesh, Wireframe and Skeleton can be used to turn on/off that part of the model. Be aware that wireframe is quite resource expensive, which is why it's off by default.

Notes:
- DDD files use CTRT as their texture format. These are encoded and swizzled and there's no code for it. You can use the remastered assets for them if needed.
- WPN files do not have bones. They use the last bones in the model they are attached to. For this tool I just added said bones in the tool for proper extraction.
- Map files are not optimized at all, they'll export the texture for every single mesh in them, which leads to a lot of duplicate data.
- The tool uses my "ModelingToolkit" as a middle ground between all models. I made my own toolkit to be able to have a lot of functionality related to modeling in a single place. It uses Assimp to import/export FBX files and HelixToolkit for the viewer.

Possible expansions if someone wants to try:
- Add the remaining formats, which are KH1 maps and the DS game files.
- Add an import function. To convert any model to another format. Note that KH1 and KH2 use vectors to store the bone positions instead of matrices. This is the main reason why I created ModelingToolkit, in it both the vectors and matrices can be stored, which makes it possible to port KH1 models to KH2 and viceversa.


kenjiuno (approved):
Great job 👍

PR #935, opened 2023-11-19
Ctd Editor improvements: round 1

thamstras:
* Ctd Editor renders text again
* You can now see/edit layouts

The CtdEncoder is still missing some special chars, how the text layout works needs more investigation, and you can't change the font used for rendering yet, but I want to get this in peoples hands before I go off down another rabbit hole...


shananas (approved):
can confirm `message/en/system/CT00000.ctd` crashes on current release after a certain amount of scrolling and this fixes it.
https://discord.com/channels/409140906625728532/831368822786621520/1175862185965129890

PR #941, opened 2023-11-25
Documentation of KH2 SCDs & 12soundinfo.bar.

Zurphing:
Some more comprehensive documentation on the SCD format for KH2. The knowledge here is probably identical for the other games that use SCD, but this is untested.

This updated SCD documentation includes some info I haven't found anywhere else, like the Sound Mapping Table that maps sound effects to ID's to be used by the game, as well as controls the volume of each sound, as well as an int in the SoundEntry table that controls how long audio should play for.

Also includes documentation on 12soundinfo.bar, this is a file that controls various ambient sound effects in KH2's maps like footsteps and environmental sounds on a per-room basis.


Vladabdf (approved):
Fixed a couple formatting errors, but otherwise looks great! Is this ready to merge now?

Zurphing:
Yep, its ready!

PR #946, opened 2023-12-06
Lua backend integration

shananas:
Allows MM to install the latest version of Lua Backend for use with Panacea and optionally configure LuaBackend.toml to run scripts from the mod manager if another file path does not exist already in the toml that ends with `mod/selected game/scripts` to prevent multiple MMs somehow being setup.

image

image

Not a huge fan of how I did everything please offer any improvements.


Vladabdf (approved):
Very well done. Thanks for this

PR #952, opened 2023-12-17
Dark Mode

shananas:
Changes Mod manager to use grey as the background and text to be white.

image


Vladabdf (approved):
You read my mind, didn't you?

PR #979, opened 2024-01-22
Adding build confirmation window

kenjiuno:
2024-01-22_20h21_18


shananas:
As the intention of the `mod` folder is to be built through openkh I am very against this window popping up every time someone builds. I'd be fine if it was a toggle say under settings that determined if it deleted mods so there's not a whole extra window to deal with every time someone builds (which for randomizers currently only kh2 which using ROM edits is very often due to changing seeds/installed mods every seed).
Also as MainMemory added a `dev_path` variable to panacea for those who wish to work directly in a folder without the need to ever `build` their mods this isn't needed but if it was its own toggle (under settings is probably the most fitting) and not an extra window id be fine with having both.

PR #981, opened 2024-01-23
Mm DDD support

shananas:
also changes panacea from looking for `ddd` to `kh3d` since thats what the pkgs are named.
image

image
image

image
image

image

image

image

image


kenjiuno:
Thanks for contributing of support code of 2.8!
Can I merge this now?

PR #1019, opened 2024-04-12
Fix panacea being unable to be built in Debug Confiiguration

shananas:
I also want to ask why by default Panacea is set to not be build under any Configuration and the user needs to check it i any of the 4 options (.NET Debug/Release and Debug/Release)
Should Panacea not be enabled to be built under Debug and Release. IF so can add to this or make another PR for that.


kenjiuno:
> I also want to ask why by default Panacea is set to not be build under any Configuration and the user needs to check it i any of the 4 options (.NET Debug/Release and Debug/Release)

This limitation (`.NET Debug` and `.NET Release` won't build Panacea) comes from our CI (Continuous Integration) configuration built on GitHub Actions.
`msbuild` command can build all type of projects. While `dotnet` command can build only .NET projects.
It needs to be fixed in the future...

kenjiuno (approved):

kenjiuno:
> This limitation (`.NET Debug` and `.NET Release` won't build Panacea) comes from our CI (Continuous Integration) configuration built on GitHub Actions.

I misunderstood. OP mentioned about Panacea won't be built on any configurations including `Debug` and `Release`.

image

shananas:
Yeah my bad i shouldve listed all 4 seperately (.NET Build, .NET Debug, Build, and Debug)
Instead i was lazy and combined them

PR #1027, opened 2024-04-17
Add logging on launcher error especially without `Launch via Epic Games` option checked.

kenjiuno:
The detail error message of launcher error is displayed and logged.

displayed:

2024-04-17_11h36_43

logged into `OpenKh.Tools.ModsManager.log`:

2024-04-17_11h34_04

Be careful! the log file is always cleared whenever we start the ModsManager.


Vladabdf (approved):
Hopefully this will be helpful with the seemingly random times that the Mod Manager breaks.

PR #1034, opened 2024-05-07
Added listpatch support for LVPM

kareking1:
Made Lvpm listpatchable. This solution should be applicable for all files that don't have an ID. Also had to include a new class in BaseTable to account for tables that only have 1 int as a header and not 2 ints...


kareking1:
Correction: The new class is for when there is no header at all

shananas (changes requested):
You also need to write a test for this in OpenKh.Tests/Patcher/PatcherTests.cs
and write documentation in docs/tool/GUI.ModsManager

kareking1:
> You also need to write a test for this in OpenKh.Tests/Patcher/PatcherTests.cs and write documentation in docs/tool/GUI.ModsManager/creatingMods.md

I added a working test and an entry to lvpm. Let me know if there's anything else

shananas:
youll need to sync changes since your PR as your adding back a removed .yml and seemingly editing some submodules. hopefuly this also fixes the build error as its unrelated to your code changes.

kareking1:
> youll need to sync changes since your PR as your adding back a removed .yml and seemingly editing some submodules. hopefuly this also fixes the build error as its unrelated to your code changes.

I'm not sure what to do about the submodules. I deleted the yml file, but I can't do anything about them... kinda just automatic. I can delete them using git rm, but idk if you want me to straight remove them

PR #1069, opened 2024-06-20
Fixes Run Shortcut for EGS and Steam

shananas:
Also trying new _pcReleaseLanguage based on responses in the discord the only valid folders for pks to be in are
EGS-JP will be in a folder named `jp`
EGS-Global (Everything else_ will be in a folder named `en`
Steam will be in a folder named `dt` (regardless of jp version or not)


Vladabdf:
Let me know when this is ready and it shall be merged, and hopefully everything is finally ready to go. 🫡

shananas:
its ready i testef it before opening the PR.
Cant really test all extractions since i only have steam version but itll become obvious quickly if i need to revert this change

kenjiuno (approved):
I have confirmed that this works on both EGS and Steam versions.
I'm using the Japanese version so it needed tweakness `pcReleaseLanguage: jp` on `mods-manager.yml`.

PR #1085, opened 2024-07-21
KH2 Object Editor - Viewer, Triggers, Collisions, Effect textures and QOL update

osdanova:
## Viewer

Changed the viewer to use ModelingToolkit's viewport (Uses Helix)

- Animations run at 15 fps (10 fps if model has over 250 bones)
- Animation slider added
- Bounding Box option added
- Keyboard controls addded (Space, left arrow, right arrow). Mouse must be over the 3D viewport
- Frame start and end are now taken from the motion metadata (Earlier it was based on actual keyframes and motions for some reason have keyframes before and after their start/end)

Viewer

## Triggers

The trigger lists have been updated both in Object Editor and the Docs.

Triggers
Trigger Docs

## Collisions

Collisions are now displayed when they are selected. Note: Due to how the viewer works this is incompatible with the auto collisions option. Note: If collision data was changed, it needs to be saved and the viewport reloaded to see the changes.

Collision

## Effect Textures

Effect textures are now correctly displayed. Can be exported and replaced with images that have the same width, height and color count (Color count can be lower)

ParticleEffects

## DPD export/ import

DPD files can now be exported and imported.

DpdExportImport

## Texture Animations colors

Texture animations are now shown with the parent image's color palette.

TexAnim

## QOL and Backend

- When searching in Kh2 Object Editor's files and motions enter can be pressed to apply the filters
- Bar class updated to accept dummies that point to "-1" (This is done by the BinaryArchive class)
- The Viewer in KH2 Object Editor has been moved to a static service for ease of use

## Dev Notes

- DPD textures can't be imported from scratch because they follow an order in memory that I can't figure out. Same with their actual placement in memory.


Oathseeker:
This all looks really cool, thanks for your hard work on this Osdanova, Ill try later tonight to take a look at this.

Oathseeker (approved):

PR #1089, opened 2024-07-24
Allow users to skip extracting remastered assets to save about 80% of the space per extracted game.

shananas:
Also prevent mod manager from copying a non existant file making a 0KB resultant file which crashes the game. (If the PR title is unwanted will remake this fix by itself as a standalone bugfix.

Most of the time this might lead to a simple mistextured model which isnt that big of a deal and imo some people who are very low on hard drive space would say it is worth it. But with mods like KH2 Randos `GoA ROM` there becomes a not small visual bug due to remastered textures not being mapped correctly. So im opening this PR as a draft for opinions. I do believe for those with < 50,100 or so GB and who dont care to mod this will be worth it but ill post 2 images showing how bad the GoA looks with GoA ROM since it uses internal textures.

image

image

was also testing with my lion sora mod and the only issue there is since i also use `type: internal` for my remastered images since i dont edit them the hud icon uses a pax image that lines up in the same slot lion soras hud is. this was my afformentioned itll only look a little off and is easily worth it. then loading into the GoA shows a much larger issue that could easily be a gameplay issue as portals dont accurately show what world they are going to. This could be fixed by the mod itself or better yet by fixing the base remastered texture linking issue


shananas:
Photos of the addition to the game data extraction page for PC only and also the new smaller size. (Not origianl size but smaller than current size)

image

image

shananas:
I know I keep editing this but I kept realizing that my bugfix to prevent 0KB files from either just being created and placed in `mod/etc` or creating a 0KB file there then trying to modify a supfile or part of a supfile kept missing a possiblity but I believe I finally covered each one and this should only prevent the 0KB file bug while allowing any valid use to happen. (This bugfix is needed regardless of my skip remastered PR also apart of this) Since I have recieved no thoughts against this and still believe it is very worth it both the `Skip Remastered` and Bugfix in this PR are ready to be reviewed and merged.

shananas:
Showing added region select from most recent commit. Chose not to expand the height of the window like i did previously whenever adding to the extract game data page for PC. Instead just made the page scrollable

image

image

PR #1090, opened 2024-07-26
Add "bbsarc" patch mode

thamstras:
Adds "bbsarc" as a valid mode in mod.ymls. Should work just like binarc.


thamstras:
[arc-patch-example.zip](https://github.com/user-attachments/files/16432463/arc-patch-example.zip)
Example mod to prove this works.

shananas (changes requested):
If it works like binarc it looks like you need to write two test in PatcherTests.cs. can refrence/~~copy~~ from the two test functions starting on line 72.
Also need to write documentation for it in docs/tool/GUI.ModsManager/creatingMods.md

shananas (approved):
Didnt test anything but the changes made are what I requested

PR #1101, opened 2024-08-19
Update extraction time msg.

shananas:
The orignal message was mean for an ISO extraction. Extraction PC version is longer due to remastered assets and also there are 5 supported games so the PC version now has its own message instead and the old one only shows for emulator.

PC Extraction page msg
image

Old msg for an ISO extraction
image


Delta-473 (approved):
LGTM

PR #1109, opened 2024-09-07
Steam API trick

shananas:
Allow mod manager to launch each of the 5 supported games directly when using steam.
If the game is installed on steam and you make a text file in the game install folder named `steam_appid.txt` with the contents being the appid of the launcher you can launch the games exes directly.

contents of the 1.5 2.5 .txt

image

contents of the 2.8 .txt

image

for any curious reader this does also work for 0.2 but it needs to be next to the 0.2 exe which is in a folder further down the heiarchy for my personal install its inh `D:\SteamLibrary\steamapps\common\KINGDOM HEARTS HD 2.8 Final Chapter Prologue\KINGDOM HEARTS 0.2 Birth by Sleep\Binaries\Win64` uses the same appid as the 2.8 one for DDD its just the 0.2's exe is in a different location

### DIFF
--- a/OpenKh.Tools.ModsManager/ViewModels/SetupWizardViewModel.cs
+++ b/OpenKh.Tools.ModsManager/ViewModels/SetupWizardViewModel.cs
@@ -78,12 +78,21 @@ namespace OpenKh.Tools.ModsManager.ViewModels
                 OnPropertyChanged();
             }
         }
+        private Xceed.Wpf.Toolkit.WizardPage _wizardPageAfterLuaBackend;
+        public Xceed.Wpf.Toolkit.WizardPage WizardPageAfterLuaBackend
+        {
+            get => _wizardPageAfterLuaBackend;
+            private set
+            {
+                _wizardPageAfterLuaBackend = value;
+                OnPropertyChanged();
+            }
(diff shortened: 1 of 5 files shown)


shananas:
New setup wizard page only shows if Steam PC Version is selected

image

OpenKh.Tools.ModsManager/ViewModels/SetupWizardViewModel.cs
```diff
@@ -78,12 +78,21 @@ private set
...
+            get => _wizardPageAfterLuaBackend;
+            private set
+            {
+                _wizardPageAfterLuaBackend = value;
+                OnPropertyChanged();
+            }
+        }
         public Xceed.Wpf.Toolkit.WizardPage PageIsoSelection { get; internal set; }
         public Xceed.Wpf.Toolkit.WizardPage PageEosInstall { get; internal set; }
-        public Xceed.Wpf.Toolkit.WizardPage PageEosConfig { get; internal set; }
```
Delta-473, line 83: Where these 2 lines that are deleted unused?

Delta-473 (approved):

PR #1118, opened 2024-09-29
Fix bug MapGen may generate a collision plane having (NaN, NaN, NaN, NaN)

kenjiuno:
Oathseeker reported that a problematic map generated by MapGen will cause at least 2 problems:

- This causes an indeterminate crash on the KH2 PC with a playing on the problematic map loaded.
- `Kh2MapStudio` will crash if we turn on the `Show map collisions` option after loading the problematic map.

Fix bug MapGen may generate a collision plane having (NaN, NaN, NaN, NaN)

> Co3[1134]   0   0 PolyCo4( 825, 826, 810,  -1) PlaneCo5(913) Co6(968) Co7(  0)
> 
> Co4[825] X:2200 Y:0.0003623899 Z:-2400 W:1
> Co4[826] X:2100 Y:0.0003623899 Z:-2400 W:1
> Co4[810] X:2050 Y:0.0003623899 Z:-2400 W:1
> 
> Co5[913] Normal:X:NaN Y:NaN Z:NaN D:NaN

According to Bing Copilot, the plane cannot be built from these 3 3D coordinates. Thus skip the invalid plane instead of including it.

2024-09-30_01h42_16


Oathseeker (approved):
I can confirm this fixes the issue with the collision generated in a custom map, thanks Kenji!

PR #1120, opened 2024-10-23
Add ScrollViewer to each setup wizard window 

shananas:
for users who increase text size through accessibility options.


Oathseeker (approved):
My name is oathseeker and I approve this message.

PR #1136, opened 2025-01-13
Move `panacea_settings.txt` file creation

shananas:
Move `panacea_settings.txt` creation into the Install/Remove Panacea Methods instead of clicking finish on the Mod Manager Wizard


shananas:
Also Fix incorrect warning

When doing an internal copy if said file was missing it would incorrectly state the assetFile.name was missing whe in reality the assetFile.Source[0].Name was missing

shananas:
Also Also
Make default postion middle of a 1920 x 1080 screen

Numbers are low enough it wont be off the screen for anyone but also wont be in the top right as the default. Still saves where you moved it to and will load it there in the future

TopazTK (approved):
LGTM!

TopazTK:
This PR has caused a regression and now Mod Manager fails to create the settings file altogether.
Please see issue #1140.

TopazTK:
Regression fixed as of #1141

PR #1138, opened 2025-01-16
Better Detect Install for Steam

shananas:
If the user had a folder with the same name as where 1.5 2.5 or 2.8 were installed it would say it is a valid install even if the folder is empty. this checks that .dll that lets you press the next button exists in that folder first.


shananas:
Detect EGS install also checks for the .dll that lets the user press next. Untested since i dont own EGS version.
Just added an if statement to check for the dll in the found game path should work fine

shananas:
Lua Backend code updates and crash prevention

TopazTK:
https://media.tenor.com/lztbLMEBDs4AAAAM/sonic-boom-knuckles.gif

TopazTK (approved):

Delta-473:
LGTM

Issue #1140, opened 2025-01-23
[REGRESSION] - Panacea Settings file is not created on setup!


TopazTK:
No need for a lot of details here, The PR #1136 has caused a regression in OpenKh.Tools.ModsManager that causes **panacea_settings.txt** to ***NOT*** be created at any time on the wizard. This causes Panacea to not do it's primary function of intercepting and loading files dynamically.

TopazTK:
Fixed as of #1141, closing.

PR #1141, opened 2025-01-23
Fix panacea_settings.txt

shananas:
being unable to be created on first time setup and unable to be deleted (when trying to delete and error/crash would happen)


shananas:
fixes #1140 and yes i tested this from a fresh version and one after completing the setup wizard

TopazTK:
Based.

TopazTK (approved):
image

PR #1143, opened 2025-01-24
More try catch to prevent setup wizard from crashing.

shananas:
if file is being used by another program with exclusive access or if it doesnt exist etc


shananas:
Auto detect Windows vs Linux.

And name both panacea and Lua Backend depending on the users OS

TopazTK (approved):
LGTM!

Issue #900, opened 2023-09-29
[BUG] - Mods breaking final form graphics in KH2


tryashtar:
**OpenKH Build**
708

**OpenKH Program**
Mod Manager

**Describe the Bug**
When applying certain mods, the graphics of the Final Form aerial finisher are broken. I reported this here, but was asked to report here instead: https://github.com/TopazTK/KH-ReFined/issues/133

**To Reproduce**
1. Apply [this mod](https://github.com/KH-ReFined/KH2-MAIN) to KH2 and build
2. Load the game and enter final form
3. Perform a full aerial combo ending with the finisher

**Expected behavior**
Vanilla finisher graphics (correct):
https://github.com/TopazTK/KH-ReFined/assets/26075577/af33a8dc-1133-408a-bf9a-6e6c8b5d2ffb
![image](https://github.com/TopazTK/KH-ReFined/assets/26075577/e813fbf1-358f-404e-9b54-d4b7ea1a3ef4)
![image](https://github.com/TopazTK/KH-ReFined/assets/26075577/201e6324-8390-4b80-b59a-f5ff75aaf5f3)

**Screenshots**
Modded finisher graphics (incorrect):
https://github.com/TopazTK/KH-ReFined/assets/26075577/fbe022f8-1152-4c05-9f44-462548c87ba3
![image](https://github.com/TopazTK/KH-ReFined/assets/26075577/0c784620-e39e-4811-90cb-bc888fdb9e85)
![image](https://github.com/TopazTK/KH-ReFined/assets/26075577/b87009df-4115-44cc-ba1f-0cd55cefb913)

Vladabdf:
I have replicated the issue. It only happens in "standard" worlds. The effects play correctly in Timeless River.

I'm not sure it's related to OpenKH though, unless somehow #897 messed things up. I can't imagine any other pull request in recent history would be responsible, as I update my mod manager every time there is an update and did not have this issue as well until 3-4 days ago.

Vladabdf:
@shananas Any ideas? Have you seen this error?

TopazTK:
> I have replicated the issue. It only happens in "standard" worlds. The effects play correctly in Timeless River.
> 
[...]

Issue #571, opened 2022-02-25
[BUG] - False linking of HD assets to their SD counterparts on duplicates with differing names.


TopazTK:
**OpenKH Build**
Build: [20220225.3](https://github.com/Xeeynamo/OpenKh/releases/tag/release-377)

**OpenKH Program**
OpenKH.Command.IdxImg

**Describe the Bug**
Two subjects:
- P_EX020.a.us - Exists in the game.
- P_EX020.a.jp - Does NOT exist in the game.

When the former is patched, the program finds 7 HD assets to link, so all is dandy.
However, when the latter is patched in, the program finds 8 SD assets to link to the HD assets, which there are only 7.
This causes the HD assets to get linked in an incorrect matter, shifting all of the effects used: breaking them.

**To Reproduce**

_I have given a **.a.jp** file as an example in the report, but I will give a **NEW.a.us** file to reproduce this, since previewing **.a.jp** files are a pain._

1. Extract **P_EX020.a.us** from **kh2_sixth**
2. Patch **P_EX020.a.us** as **P_EX020_NEW.a.us** (And all the co-files that go along with it. Like the MSET and the MDLX)
3. Preview the file by changing **P_EX020** to **P_EX020_NEW** in the **00objentry.bin** (Preferably from RAM)
4. Observe.

**Expected behavior**
The files act 1:1 to one another. No messiness, no tricks, no compromises.

**Screenshots**
![image](https://user-images.githubusercontent.com/95656963/155760346-4242d8d6-47e2-4113-a71f-cfceb7ebac2b.png)
![image](https://user-images.githubusercontent.com/95656963/155760381-10e555c5-4013-4223-ac80-44fdf2d352fc.png)
![image](https://user-images.githubusercontent.com/95656963/155760395-67edc6fa-7e32-430e-8c66-7a55f1206ce7.png)

**Additional context**
With the context of all **.a.us** files copied and patched in as **.a.jp**, and the game modified to seek the latter instead of the former, the following are true:

Screenshot 1: P_EX020.a.jp reports 8 SD files found. This should be 7.
Screenshot 2: Save Point assets are out of order and so it appears broken. Donald's face is still SD.
Screenshot 3: All in battle PAX are out of order, so it appears buggy.

TopazTK:
I can think of 3 ways to fix this issue:
- Allow extracting/patching with unstripped files for maximum compatibility.
[...]

PR #1151, opened 2025-02-26
[Panacea] Fix Remastered texture sorting to properly order & combine textures for DPDs

Zurphing:
Addresses #571 and #900.

This changes the way Panacea handles texture sorting when dealing with newly patched PAX files. 

Before this, Panacea would look for textures in newly patched PAX files in order of offsets found. So if it found Offset 0, 1, and 2 for DPD 0, then it would expect HD textures to be applied in that same order. However, extracted copies of the game would not have the textures in this order, and instead, might have textures in the order of Texture 2, 1, 0, which necessitated trial and error in order to try and order textures according to how Panacea expected the order. In addition to this, Panacea could sometimes get the texture count wrong due to using a different value to determine if two or more textures should be combined.

For example, this is how Marluxias Absent Silhouette portal when extracted from the game would look.
image
However, to get Panacea to link HD textures, you would need to order it like this:
image

This is due to the game using the first 2-byte value in each DPD Textures header to determine Texture Order. This value is likely how the devs chose to sort PAX textures, and when two textures share this value, they need to be combined into one "combo texture." 

With this change, you can patch in a new PAX file, copy and paste the textures from the remastered folder, and it should link without any manual adjustment.

Leaving this as a draft for now, this change would break existing mods that patch in PAX files and I'd like there to be enough of a heads up before it gets pushed. I might end up bundling in a similar sorting method for TEXA textures like Sora's facial expressions in order to completely remove discrepancies between how the game is extracted and how textures should be applied, which would make all of the breaking changes contained to just one update.

### DIFF
--- a/OpenKh.Research.Panacea/Panacea.cpp
+++ b/OpenKh.Research.Panacea/Panacea.cpp
@@ -435,40 +435,25 @@ void GetTM2Offsets(void* addr, int baseoff, std::vector<int>& entries)
     }
 }
 
-void GetDPDOffsets(void* addr, int baseoff, std::vector<int>& entries) //Process texture offsets from given memory address, 'addr' and stores results in entries.
+void GetDPDOffsets(void* addr, int baseoff, std::vector<int>& entries)
 {
     int* off = (int*)addr;
     if (*off++ == 0x96) //Magic number check.
     {
-        off += *off + 1; //Read the next integer *off as a "SkipCount"
+        off += *off + 1;
         int texcnt = *off++; //Read number of textures
-        printf("Texture count: %p\n", texcnt); // Verify texture count
 
         std::vector<std::pair<short, int>> textures; // To store texture offsets and their first two bytes. First entry stores firstTwoBytes, second stores texture offset
 
         for (int t = 0; t < texcnt; ++t) //Iterates over textures.
         {
             int texoff = *off++; //Reads texoff from off.
-            printf("Processing texture %d at offset: %p\n", t, texoff);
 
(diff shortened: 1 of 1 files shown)


shananas:
I support this and combining it with TEXAs all in one. Note this should also be done for the Patcher too otherwise Panacea and Patching will differ between the same mods. So id like to see this fixed for all (PAX and TEXAS for Panacea and Patching)

TopazTK:
> I support this and combining it with TEXAs all in one. Note this should also be done for the Patcher too otherwise Panacea and Patching will differ between the same mods. So id like to see this fixed for all (PAX and TEXAS for Panacea and Patching)

The last I have checked, Patching did not have this issue. Panacea has this issue due to the way it replaces file routines.
Can't really have this problem if you are just replacing the RAW packaged files themselves.

shananas:
> > I support this and combining it with TEXAs all in one. Note this should also be done for the Patcher too otherwise Panacea and Patching will differ between the same mods. So id like to see this fixed for all (PAX and TEXAS for Panacea and Patching)
> 
> The last I have checked, Patching did not have this issue. Panacea has this issue due to the way it replaces file routines.
> Can't really have this problem if you are just replacing the RAW packaged files themselves.

Guess I'm misremembering then. Yeah just Parity with all would be nice if this is only a Panacea issue then ok.

MainMemory (approved):
Seems good to me.

MainMemory (approved):
This is also good!

shananas (approved):

PR #1157, opened 2025-03-05
Command Line Tool for Converting Between KH Animation Formats - KHAnimationConverter

Some1fromthedark:
Adds a new command line tool for converting between the animation formats used by the various Kingdom Hearts games.

Currently only supports converting from PAM (BBS) to ANBs (2/FM).

This tool is based on Disgustor's PAM Converter command line tool, but has been ported to use the OpenKH libraries.
I also included the Unwrap Angles fix so that the converted animations no longer flicker.


TopazTK:
Godsend, but please try to fix the build fail. If can't at this moment, mark this as a draft for the time being.

Some1fromthedark:
Build is successful after removing all the implicit stuff. Let me know if anything else is necessary before merging.

TopazTK:
You can use LangVersion 12 with NET6 btw.
But due regulations we have, Imma ask you to implement at least 2 unit tests.

Some1fromthedark:
Added Unit Tests for testing the conversion of PAM -> ANBs and the exporting of ANB files.

TopazTK (approved):
Based!

PR #1159, opened 2025-03-12
DRAFT: Image viewer make tm2s

thamstras:
For some reason the image viewer can't export as .tm2.
This fixes that.
It also merges the `IImage` and `IImageRead` interfaces because there's no practical reason to separate them and it just confuses people.
I wrote this a year ago and never pushed it for some reason so it probably breaks things.


kenjiuno:
Hi
If you don't mind, I want to create a new pr with minimal changes (without removing `IImage` and without adding `ClutFormat`)

TopazTK:
This PR is over a year old, is it still valid and/or relevant?

PR #1162, opened 2025-03-17
Kh2ObjectEditor - Collision ComboBox fix

osdanova:
Dis kindly found another bug in my flawless code so I could improve upon perfection.


osdanova:
Oh whoops, thanks for checking it out. I've uploaded the fix.

PR #1163, opened 2025-03-21
OpenKH ModManager: Add online mod repository browser and downloader with thread-safe image loading

Yokimitsuro:
This commit introduces a new online mod repository browser and downloader feature with several improvements:

- Adds functionality to browse, view and download mods from online repositories
- Implements a new UI section for browsing downloadable mods by game
- Creates a downloadable mods list populated from JSON configuration
- Implements thread-safe image loading using Dispatcher to prevent UI threading errors
- Makes URL retrieval methods public and accessible for proper image display
- Adds proper exception handling during image loading process
- Updates the available mods list with real and current repositories
- Optimizes the DownloadableModViewModel constructor to properly initialize properties
- **JSON URL** in DownloadableModsService.cs Line 22: 
`private const string DEFAULT_MODS_JSON_URL = "https://raw.githubusercontent.com/OpenKH/mods-manager-feed/main/downloadable-mods.json";`

These changes allow users to discover and install mods directly from the application interface without the DependencySource errors that were previously occurring.
image
image

### DIFF
--- a/OpenKh.Tools.ModsManager/ViewModels/SetupWizardViewModel.cs
+++ b/OpenKh.Tools.ModsManager/ViewModels/SetupWizardViewModel.cs
@@ -575,13 +575,11 @@ namespace OpenKh.Tools.ModsManager.ViewModels
             {
                 if (PcReleaseLocation != null && GameCollection == 0)
                 {
-                    return (File.Exists(Path.Combine(PcReleaseLocation, "LuaBackend.dll")) || File.Exists(Path.Combine(PcReleaseLocation, "dinput8.dll"))) &&
-                        File.Exists(Path.Combine(PcReleaseLocation, "LuaBackend.toml"));
+                    return File.Exists(Path.Combine(PcReleaseLocation, "LuaBackend.dll")) && File.Exists(Path.Combine(PcReleaseLocation, "LuaBackend.toml"));
                 }
                 else if (PcReleaseLocationKH3D != null && GameCollection == 1)
                 {
-                    return (File.Exists(Path.Combine(PcReleaseLocationKH3D, "LuaBackend.dll")) || File.Exists(Path.Combine(PcReleaseLocationKH3D, "dinput8.dll"))) &&
-                        File.Exists(Path.Combine(PcReleaseLocationKH3D, "LuaBackend.toml"));
+                    return File.Exists(Path.Combine(PcReleaseLocationKH3D, "LuaBackend.dll")) && File.Exists(Path.Combine(PcReleaseLocationKH3D, "LuaBackend.toml"));
                 }
                 else
                     return false;
@@ -1204,14 +1202,7 @@ namespace OpenKh.Tools.ModsManager.ViewModels
                         {
                             if (File.Exists(Path.Combine(TempExtractionLocation, "DBGHELP.dll")))
                             {
-                                if (Process.GetProcessesByName("winlogon").Length > 0)
-                                {
-                                    File.Move(Path.Combine(TempExtractionLocation, "DBGHELP.dll"), Path.Combine(DestinationCollection, "LuaBackend.dll"), true);
-                                }
-                                else
-                                {
-                                    File.Move(Path.Combine(TempExtractionLocation, "DBGHELP.dll"), Path.Combine(DestinationCollection, "dinput8.dll"), true);
-                                }
+                                File.Move(Path.Combine(TempExtractionLocation, "DBGHELP.dll"), Path.Combine(DestinationCollection, "LuaBackend.dll"), true);
                             }
                             if (File.Exists(Path.Combine(TempExtractionLocation, "LuaBackend.toml")))
                             {
@@ -1538,13 +1529,11 @@ namespace OpenKh.Tools.ModsManager.ViewModels
                         if (GameCollection == 0)
                         {
                             File.Delete(Path.Combine(PcReleaseLocation, "LuaBackend.dll"));
-                            File.Delete(Path.Combine(PcReleaseLocation, "dinput8.dll"));
                             File.Delete(Path.Combine(PcReleaseLocation, "LuaBackend.toml"));
                         }
                         else if (GameCollection == 1)
                         {
                             File.Delete(Path.Combine(PcReleaseLocationKH3D, "LuaBackend.dll"));
-                            File.Delete(Path.Combine(PcReleaseLocation, "dinput8.dll"));
                             File.Delete(Path.Combine(PcReleaseLocationKH3D, "LuaBackend.toml"));
                         }
                         OnPropertyChanged(nameof(IsLuaBackendInstalled));


shananas:
Only taking a quick look now not a review. but looking at the image this Downloadable Mods section should be its own window not apart of the main view. Some option under `Mods` that can pull this up in a separate window (can just be a copy of main window but only to show downloadable mods and if they are installed). Having this massive window as the new default especially when its not every time you run MM that you want to see downloadable mods.

Prior to an actual review make this its won window (id suggest under the `Mods` option to open it) that a user can pull up to browse installable mods.

Yokimitsuro:
I completely agree with your recommendation. Pulling repositories directly from GitHub using tags like `openkh-mod` or `openkh-mods` is definitely a cleaner and more dynamic solution, reducing the need for constant manual updates.

The idea of implementing a blacklist to handle problematic repositories is also excellent. This gives us flexibility and control, helping to maintain quality standards without excessive micromanagement.

Regarding the UI design, I fully concur with both you and @Shananas—a separate, dedicated window accessible via a clearly visible **"Browse Mods from GitHub"** button would be optimal. We can avoid cluttering the main interface while ensuring the feature remains discoverable and convenient.

### Implementation options:

**1. Using the GitHub API:**

- **Pros:**
  - ✅ Clean, structured, officially supported by GitHub.
  - ✅ Less prone to breaking changes.

- **Cons:**
  - ❌ GitHub API rate limits requests to **60 per hour** for unauthenticated users.
  - ❌ Authenticated users get higher rate limits (**5,000 per hour**). We would likely need users to provide a GitHub token or implement OAuth to avoid hitting these limits.

**2. Scraping GitHub directly (HTML parsing):**

- **Pros:**
  - ✅ No rate limit constraints.

- **Cons:**
  - ❌ Fragile; if GitHub changes their page structure, scraping logic will need frequent maintenance.
  - ❌ Could be slower or potentially violate GitHub's Terms of Service.

Given these trade-offs, I'm inclined toward using the API option to ensure robustness and reliability. 

**Do you agree with this, or would you prefer we explore scraping further?**

Additionally, to categorize and filter mods accurately by game (`KH1`, `KH2`, `BBS`, `ReCoM`, `KH3D`, etc.), we'd need to enforce that each mod repository explicitly provides clear categorization, either within the existing `mod.yml` file or through another standardized method. This would enable accurate filtering when users browse mods.

Does this approach sound good to you? Let me know your thoughts or if you'd like to consider anything else.

shananas:
My category message was posted to the implementation as is. Since you have a json with categories if we are switching away from that that part of the suggestion may need to be reconsidered. Also iirc we technically already have a blacklist. I forget which file it's in but mod manager does pull from one within a separate openkh repo it just currently has no mods that are blacklisted.

Yokimitsuro:
something like this?


_Work In Progress_
image
image

TopazTK:
I don't like the Mod Manager itself having this functionality for a multitude of reasons, some mods requiring external steps to install for one (such as Re:Fined and Critical Mix). Solid idea, but this would need to be done in an external way.

I am rejecting this PR, if anyone would object they can do so.

TheSoraHD:
> I don't like the Mod Manager itself having this functionality for a multitude of reasons, some mods requiring external steps to install for one (such as Re:Fined and Critical Mix). Solid idea, but this would need to be done in an external way.
> 
> I am rejecting this PR, if anyone would object they can do so.

In fact, I think it is NECESSARY to have an internal mod search tool. Sometimes, finding a specific mod outside of NexusMods can be tricky, and this could be a great solution to help end users install more mods than ever, in a centralised way, without the need to search all over the internet.

I don't think mods that require external installation steps would be severely affected (like Re:Fined/Critical Mix); it would be as simple as indicating in the respective mod.yml (either through the description or a new boolean field) the need for extra steps to make them work properly.

TopazTK:
Let me elaborate some points just so I am clear:

- GitHub: As you have mentioned in the first place, the GitHub API doesn't really play nice to these kinds of implementations sadly, without dire consequences and/or unintuitive workarounds. Without severe optimizations, we are surely to surpass 60 API requests per hour, and we can't expect people to provide an OAuth key when the Mod Manager alone manages to scare most people as-is.
- External Installation: Some mods which tend to overhaul the game (such as Critical Mix or Re:Fined), some mods which have a way of altering how it's played (such as the Randomizers) and some that just do shit out of the blue (such as Aqua/Terra Ally) may or do require external meddling and fiddling to get completely working. Just a simple mod parse from the Mod Manager will most likely the way to be introduced to those kinds of mods, and when people inevitably install them incompletely, they will either flood the OpenKH server or the mod's own server to complain that their installation does not work. To anyone that may see this point as moot I would like to remind you that Re:Fined and OpenKH get these kinds of complains about their own and/or unrelated stuff **on a daily basis**.
- Out-of-Date Mods: We can't force people to start adding mod version tags on the GitHub when it's a wild-west kind of a place for mod uploads. I have sounded concerns many times that GitHub would be a limiting factor on these kinds of updates when the time comes, and it shows. As there is no moderation on GitHub, people will **not** use the tags that they should and there is absolutely nothing we can do to make them do so. We can tell them to use "openkh-mod" if they want their mods to be shown, and provide a game version, but either they will not do so, or will do so incorrectly, stating a mod is for Steam v1.0.0.10 when in reality it's for EGS v1.0.0.8 and hasn't been updated just so it can appear on the Mod Manager download page.
- Plagiarizing and Stolen Work: As soon as we put an internal download on OpenKH, we are solely responsible to moderate each and every one of the mods that appear in it, as we are directly or indirectly promoting each and every work. We do have a blacklist system, yes, but such systems on FOSS projects and with a GitHub access portal are not only frowned upon, but they may also even be against certain ToS. Even if they are not, there are currently THOUSANDS of OpenKH mods floating around. Some private, some public, some NSFW, some stolen. The OpenKH has a moderation team of like 10 people with a programmer team of like, 5? At best? And each and every one of these people have lives. We can't have an **on-approve** system because GitHub, and we cannot afford to be held responsible if a "Nude Kairi" or a "Nude Namine" mod makes the rounds, or if a private "Big-Ass Purple Bat" mod **accidentally** appears on the mod manager. Huge ass responsibility that we cannot take or uphold.

TopazTK:
> I don't think mods that require external installation steps would be severely affected (like Re:Fined/Critical Mix); it would be as simple as indicating in the respective mod.yml (either through the description or a new boolean field) the need for extra steps to make them work properly.

Svenchu, you are a translator in the Re:Fined project, you out of all people should 100% be aware that people can't install Re:Fined correctly even with baby-step instructions. They will **NOT** see the description of the **mod.yml** file.

TopazTK:
I retract all my previous statements. Do whatever.

shananas:
Im unsure github API is the way to go and half agree with topaz's removed statements. so i have this proposal
1) keep the json but not in this repo probably move it to the one i linked earlier or the org can make another just for this json.
  a) it prevents mods that are not a simple click to install (that may be tagged as vald said) from showing up its a good point refined shouldnt show up in this mod browser becuase it will confuse people since the mod does nothing by itself..
  b) the json being in a different repo prevents a new release being built every time. also id expect anytime a mod is added the author is aware and ok with it being added ( i dont think this needs to be a question every time but a general hey we have this feature now can i PR to this repo so it shows up in the mod browser if the user hardly uses github or they can PR it themselves)

still havent reviewed the code but no reason to stop working on it itll be merged.
from the latest images got to fix the text being the wrong color in dark mode (ik it was a PoC) 
Then move the json i think adding it to the already exisitng repo for things like this https://github.com/OpenKH/mods-manager-feed would be best. Id be glad to accept a PR with a couple mods into that repo to show its working without the json in this repo.

Yokimitsuro:
https://github.com/OpenKH/mods-manager-feed/pull/1

Yokimitsuro:
This PR depends on OpenKH/XeEngine.Tools.Public#1

The ModsManager requires the RefreshIcons.xaml resource file that has been submitted as a PR to the XeEngine.Tools.Public submodule. Once that PR is merged, this PR can be updated to use the official repository version.

Currently, the submodule is using a custom branch with these changes to make the application work correctly.

shananas:
> This PR depends on [OpenKH/XeEngine.Tools.Public#1](https://github.com/OpenKH/XeEngine.Tools.Public/pull/1)
> 
> The ModsManager requires the RefreshIcons.xaml resource file that has been submitted as a PR to the XeEngine.Tools.Public submodule. Once that PR is merged, this PR can be updated to use the official repository version.
> 
> Currently, the submodule is using a custom branch with these changes to make the application work correctly.

merged

Yokimitsuro:
updated json

OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
```diff
@@ -1177,155 +1240,215 @@ private async Task FetchUpdates()
...
+                PresetList = new ObservableCollection<string>();
+
+            PresetList.Clear();
+            foreach (string presetFilePath in Directory.GetFiles(ConfigurationService.PresetPath))
+            {
+                PresetList.Add(Path.GetFileNameWithoutExtension(presetFilePath));
+            }
+        }
+
+        public async void RefreshDownloadableMods()
```
shananas, line 1394: Is this method needed anymore since the downloadable mods view is now its own window and not apart of main view model

OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
```diff
@@ -1,511 +1,518 @@
...
+        public static bool SkipRemastered
+        {
+            get => _config.SkipRemastered;
+            set
+            {
+                _config.SkipRemastered = value;
+                _config.Save(ConfigPath);
+            }
+        }
+        public static string LaunchGame
```
shananas, line 485: Is this file needed to be edited anymore since the downloadable mod view is now its own window. My thought is you can have your own dropdown in the new window to filter which game to show (or maybe show all if the user wanted). and even if the launch game is chosen for KH2 you can install a mod for say KH1 and it will install into the correct folder.

shananas: I may be wrong and it might be easier to still attach to LaunchGame if thats the case then so be it. idk what better attaching to LaunchGame or allowing ModsService.InstallMod to use a different path than the currently selected LaunchGame

OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
shananas: If the refresh method is still needed I feel it would be better in DownloadableModsService

Yokimitsuro: The change to move the view to a new window hasn't been done yet. I'm working on it — it still needs adjustments and it's not included in this PR.

This PR currently contains what I showed in the first screenshots: mod downloads within the same window, but with separate loading.

shananas: I'd prefer to wait for it to be its own window to merge this PR. Its fine if it takes another week or so. Whenever you are able to finish it. the ballooning of the main window here i think is visually undesirable enough to wait for the new window implementation to be complete.
again no rush just whenever you are able to finish that code

shananas:
Wanted to check on this. (moving the mod browser to its own separate window ETC.

Yokimitsuro:
> Is this superseded by #1187?

yes

PR #1166, opened 2025-04-15
[Mod Manager] Adjust objentry listpatching

Zurphing:
Small tweak to objentry listpatching to sort newly added entries by ID before then patching them in. 

The game requires entries in objentry to be in numerical order, and will refuse to load in entries that don't adhere to this order.
Example being, Mod A listpatches an ID of 5000, and Mod B listpatches an ID of 4562.

If Mod A has lower priority than Mod B, then ID 5000 would be appended to objentry first, followed by the id of 4562. This means that any ObjID's lower than 5000 that get built after Mod A would fail to load in-game. In order for these two mods to work together, Mod A would need to be placed at a higher priority than Mod B.

Sorting by IDs first fixes this issue, as regardless of order, 4562 would be appended first, followed by 5000.


Delta-473 (approved):
Didn't test and had to lookup if `objEntryList[objEntry.Key] = objEntry.Value;` was valid and that it wouldn't throw invalid key exceptions but it's correct. And it's a little bit faster :) so LGTM https://stackoverflow.com/questions/4245064/method-to-add-new-or-update-existing-item-in-c-sharp-dictionary

PR #1174, opened 2025-06-05
Add support for Github multi game collections

cahudson94:
Adds definitions to the mod.yml file for the following fields:
- isCollection: boolean for if the specific mod install is actually a collection (ie contains multiple games worth of info)
- collectionGames: string list of the games that have mods included
- asset.game: string to specify which game the script belongs with

This allows for the install and update of collection based mods from Github using the above tags in the mod.yml as well as the update and auto update of said installed mods. Creates a parent folder called collections on the same level as the game specific folders. And cleans up any installation of the mod that happened before it became a collection.

### DIFF
--- a/OpenKh.Tools.ModsManager/Services/ModsService.cs
+++ b/OpenKh.Tools.ModsManager/Services/ModsService.cs
@@ -76,6 +76,19 @@ namespace OpenKh.Tools.ModsManager.Services
                     if (File.Exists(Path.Combine(dir, ModMetadata)))
                         yield return authorName;
                 }
+                var collectionsPath = ConfigurationService.ModCollectionsPath;
+                foreach (var dir in Directory.GetDirectories(collectionsPath))
+                {
+                    var authorName = Path.GetFileName(dir);
+                    foreach (var subdir in Directory.GetDirectories(dir))
+                    {
+                        var repoName = Path.GetFileName(subdir);
+                        if (File.Exists(Path.Combine(subdir, ModMetadata)))
+                        {
+                            yield return $"{authorName}/{repoName}";
+                        }
+                    }
+                }
             }
         }
 
@@ -286,7 +299,7 @@ namespace OpenKh.Tools.ModsManager.Services
                 throw new ModNotValidException(repositoryName);
 
             var modPath = GetModPath(repositoryName);
-            if (Directory.Exists(modPath))
+            var collectionPath = GetCollectionPath(repositoryName);
             {
                 var errorMessage = MessageBox.Show($"A mod with the name '{repositoryName}' already exists. Do you want to overwrite the mod install.", "Warning", MessageBoxButton.YesNo, MessageBoxImage.Question, MessageBoxResult.No , MessageBoxOptions.DefaultDesktopOnly);
 
@@ -296,14 +309,19 @@ namespace OpenKh.Tools.ModsManager.Services
                         Handle(() =>
                         {
                             MainViewModel.overwriteMod = true;
-                            foreach (var filePath in Directory.GetFiles(modPath, "*", SearchOption.AllDirectories))
+                            if (Directory.Exists(modPath))
+                                CleanModFiles(modPath);
+                            else
                             {
-                                var attributes = File.GetAttributes(filePath);
-                                if (attributes.HasFlag(FileAttributes.ReadOnly))
-                                    File.SetAttributes(filePath, attributes & ~FileAttributes.ReadOnly);
+                                CleanModFiles(collectionPath);
+                                foreach (var game in new List<string> { "kh1", "kh2", "bbs", "Recom", "kh3d" })
+                                {
+                                    modPath = GetModPathGame(repositoryName, game);
+                                    if (Directory.Exists(modPath))
+                                        CleanModFiles(modPath);
+                                }
                             }
-
-                            Directory.Delete(modPath, true);
+                                
                         });
                         break;
                     case MessageBoxResult.No:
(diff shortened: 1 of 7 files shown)


TopazTK:
Good shit.

cahudson94:
Yeah, I was looking at the docs a bit. I will go ahead and do it on that branch.

Delta-473 (approved):
LGTM

OpenKh.Tools.ModsManager/Services/ModsService.cs
```diff
@@ -285,8 +298,16 @@ public static async Task InstallModFromGithub(
...
-            var modPath = GetModPath(repositoryName);
-            if (Directory.Exists(modPath))
+            var modPaths = new List<string> { 
+                GetModPath(repositoryName),
+                GetCollectionPath(repositoryName),
+                GetModPathGame(repositoryName, "kh1"),
+                GetModPathGame(repositoryName, "kh2"),
+                GetModPathGame(repositoryName, "bbs"),
+                GetModPathGame(repositoryName, "Recom"),
+                GetModPathGame(repositoryName, "kh3d"),
```
shananas, line 308: These 5 should be unnecessary unless the collections addition change something larger than I'm seeing in how GetModPath works. Before this PR its only relevant to check whatever the current `LaunchGame` is because thats where the mod is being installed. After this update youll need to check just current GetModPath and your new CollectionPath. if for some reason there was a mod not labled as a collection the user had installed under say KH1 when KH2 is the selected LaunchGame then it would falsely label that as already installed and maybe even delete the KH1 installation then install it under KH2.

OpenKh.Tools.ModsManager/Services/ModsService.cs
```diff
@@ -295,22 +316,16 @@ public static async Task InstallModFromGithub(
...
+                            ClearOldMods(repositoryName);
                         });
                         break;
                     case MessageBoxResult.No:
                         throw new ModAlreadyExistsExceptions(repositoryName);
                 }
-            }
-
+            }
+            await Task.Delay(1); // fixes a bug where folder creation and deletions collide
```
shananas, line 327: Is this only delaying 1ms is that long enough or should it be more like 40ms or something. If 1ms is enough thats fine i guess

OpenKh.Tools.ModsManager/Services/ModsService.cs
```diff
@@ -336,6 +351,8 @@ await Task.Run(() =>
 
                 Repository.Clone($"https://github.com/{repositoryName}", modPath, options);
             });
+            if (IsCollection(repositoryName))
+                await MoveToCollection(repositoryName);
```
shananas, line 355: Why are you installing it first then moving it? Can you not put an `if` inside the above run and if its a collection install it there first or do you need the mod.yml metadata first so you need to install the mod first.

OpenKh.Tools.ModsManager/Services/ModsService.cs
```diff
@@ -409,22 +426,30 @@ public static string GetModPath(string author, string repo) =>
 
         public static string GetModPath(string repositoryName) =>
             Path.Combine(ConfigurationService.ModCollectionPath, repositoryName);
+        public static string GetModPathGame(string repositoryName, string game) =>
+            Path.Combine(ConfigurationService.ModCollectionPath, "..", game, repositoryName);
```
shananas, line 430: As stated in another comment this is useless its simply doing what the above static variable is doing. hence you needing to do "..". quick example ModCollectionPath is `C'/Mod Manager/mods/kh2` you're going up a folder to `C'/Mod Manager/mods` then adding on the game of KH2 and getting back to `C'/Mod Manager/mods/kh2`

OpenKh.Tools.ModsManager/Services/ModsService.cs
cahudson94: The 1 ms has been consistently working

OpenKh.Tools.ModsManager/Services/ModsService.cs
cahudson94: These are in here for clean up purposes. In the case someone changed a mod to a collection but it had been installed for instance under bbs, but the current launch game is set to kh1, when the install goes through as a collection it would not properly clean up the now outdated bbs install files.

OpenKh.Tools.ModsManager/Services/ModsService.cs
cahudson94: Need the mod.yml.

OpenKh.Tools.ModsManager/Services/ModsService.cs
cahudson94: This is actually needed in a very specific case, this is applicable for where someone changes a collection mod back to a non collection. The currently selected launch game could differ from the intended install game in this case and as the mod is updated from the freshly pulled Github mod.yml we would want it to land in the appropriate game location. Admittedly not a SUPER likely scenario but I tried to make this as bug proof as possible.

OpenKh.Tools.ModsManager/Services/ModsService.cs
TopazTK: The 1ms wait is actually a known hack amongst the C# projects. Very time sensitive projects use it to either prevent race conditions or to improve CPU usage significantly, or both.

cahudson94: This one was a race condition ^

OpenKh.Tools.ModsManager/Services/ModsService.cs
shananas: This causes another edge case bug that needs to be fixed in itself if for some reason someone made a mod that isnt a collection but edits say KH1 + 2 then you cannot install it to both games. I'll take everything else as is but if for some reason the mod author does this the end user should not be punished for such. Otherwise this PR is ready to be merged.

shananas (approved):

PR #1177, opened 2025-06-10
Libretto Listpatch Update & Added "Index" binarc patching note

Zurphing:
Updates Libretto listpatching to be slightly more clear with less unknowns. 
```
- TalkMessageId: 4865 #Id to update.
  Type: 2 #Type of TalkMessageId.
  Contents: #Contents to update. 
    - CodeType: 0x0001
      Unknown: 0x0001
      TextId: 0x5501
```

Also adds in a note on creatingMods about the "index" binarc patching method, which can be used instead of the "name" binarc patching method to patch files with subfiles that have the same name like localset, various MSN files, etc.

### DIFF
--- a/OpenKh.Tests/Patcher/PatcherTests.cs
+++ b/OpenKh.Tests/Patcher/PatcherTests.cs
@@ -2800,7 +2800,7 @@ namespace OpenKh.Tests.Patcher
                 writer.Flush();
             });
 
-            patcher.Patch(AssetsInputDir, ModOutputDir, patch, ModInputDir);
+            patcher.Patch(AssetsInputDir, ModOutputDir, patch, ModInputDir, Tests: true);
 
             AssertFileExists(ModOutputDir, "libretto-ca.bar");
 


OpenKh.Tests/Patcher/PatcherTests.cs
```diff
@@ -2792,14 +2792,15 @@ public void ListPatchLibrettoTest()
...
+                writer.WriteLine("  Type: 0");
                 writer.WriteLine("  Contents:");
-                writer.WriteLine("    - Unknown1: 0x02000200");
+                writer.WriteLine("    - CodeType: 0x0200");
+                writer.WriteLine("      Unknown: 0x0200");
                 writer.WriteLine("      TextId: 2500");
                 writer.Flush();
             });
 
-            patcher.Patch(AssetsInputDir, ModOutputDir, patch, ModInputDir, Tests: true);
```
Delta-473: What does Tests:true do that you removed?

Zurphing: Woops, I'm not sure, but I forgot to update the PatcherTests file in my fork, should be fixed now.

PR #1185, opened 2025-06-30
Dont name LuaBackend.dll differently on Linux

shananas:


Delta-473:
What was the reason we used dinput8.dll on Linux as name?

TopazTK:
In the previous Proton versions, Panacea could not chainload DLLs. This was fixed in Proton 9.

Delta-473:
The only thing I'm worried for are what to do with the existing installs on Linux, people are going to need to remove the dll manually? And is something going to break if they don't remove the dll and the update version places a second copy of it?

Delta-473 (approved):
LGTM

PR #1189, opened 2025-07-05
Allow a mod to be installed across multiple games if its not a collection.

shananas:
This PR is very much my preference and maybe it can be chalked down to have the mod author fix their mod. But as of the Mod Collections PR you cannot install a non collection mod across multiple games. this removes that restriction.
i would like the opinion of multiple OpenKh orgs of if they want to keep it the way it is as of the PR merge earlier today or would like it to work as this PR makes it work

### DIFF
--- a/OpenKh.Tools.ModsManager/Services/ModsService.cs
+++ b/OpenKh.Tools.ModsManager/Services/ModsService.cs
@@ -298,14 +298,9 @@ namespace OpenKh.Tools.ModsManager.Services
             if (!isValidMod)
                 throw new ModNotValidException(repositoryName);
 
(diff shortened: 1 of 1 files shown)


Delta-473:
What are we going to do with this?

shananas:
> What are we going to do with this?

My preference remains that a user should be able to install a mod across multiple games even if the mod is not a collection. if a mod edits multiple games then realistically it should be made a collection. The author of the modcollection addition disagreed. so as to not prevent what in my opinion was a perfectly acceptable PR i merged it and made the change i wanted to see to the original PR per my preference.

kenjiuno (approved):

PR #1194, opened 2025-07-31
[MapGen] Include Group & floorLevel support for DOCT files, fix disableBSPCollisionBuilder2 from not applying properly.

Zurphing:
Title.

Using the disableBSPCollisionBuilder2 option now applies a new form of DOCT generation which allows you to assign specific faces into their own rendering nodes using the "group" tag on the material of said mesh. This allows you to hide and unhide the meshes using MapVisibility flags in the ARD.

This change also now allows you to flag specific material faces with a "floorLevel" value, which the game uses for shadow rendering.


Delta-473 (approved):
LTGM

PR #1195, opened 2025-08-01
Fix: Correct matrix transform order for AABB in KH2 Map Studio

Disgustor:
Rotation was coming before scale, causing weird skewing issues when rotating a bounding box. It's fixed by moving scale before rotation.


Delta-473 (approved):
Approved

PR #1197, opened 2025-08-04
Fix mod manager not opening sometimes

shananas:
If mod manager cannot open due to the log file being in use it will kill all processes of mod manager then restart to properly load the log file. (allowing the debug window to show up properly since that relies on the log)


Oathseeker (approved):
My name is Oathseeker, and I approve this message.

PR #1224, opened 2025-11-05
[Mod Manager] MainViewModel: Cleanup UpdatePanaceaSettings

Wryyyong:
## Changes
- Update `mod_path` text to use the same flattened directory path as in #1215
  - In between that PR and now, whenever the mod manager updated Panacea's settings file in situations outside of Building and Running the game, it would write the old expected `mod_path` value rather than the new one.
- Minimized duplicate code in the function
  - The only difference between the code in the conditionals was the path to the game's installation directory. So I moved all that code outside of the conditionals and made the conditionals simply assign a target variable.

### DIFF
--- a/OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
+++ b/OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
@@ -1353,7 +1353,7 @@ namespace OpenKh.Tools.ModsManager.ViewModels
                     configTargetPath = ConfigurationService.PcReleaseLocationKH3D;
                 else
                 {
-                    Log.Warn("Unable to update Panacea settings! Game installation directory is not configured or is invalid.")
+                    Log.Warn("Unable to update Panacea settings! Game installation directory is not configured or is invalid.");
 
                     return;
                 }


OpenKh.Tools.ModsManager/ViewModels/MainViewModel.cs
```diff
@@ -1345,42 +1345,34 @@ public void UpdatePanaceaSettings()
...
-                        {
-                            textToWrite += entry;
-                            break;
-                        }
-                    }
-                    textToWrite += $"\r\nshow_console={_panaceaConsoleEnabled}\r\n" +
-                        $"debug_log={_panaceaDebugLogEnabled}\r\nsound_debug={_panaceaSoundDebugEnabled}\r\n" +
-                        $"enable_cache={_panaceaCacheEnabled}\r\nquick_menu={_panaceaQuickMenuEnabled}";
-                    File.WriteAllText(panaceaSettings, textToWrite);
+                    Log.Warn("Unable to update Panacea settings! Game installation directory is not configured or is invalid.");
```
shananas, line 1356: Technically this should be a MessageBox.Show() so the user can actually see the error instead of silently failing (since the Log Window is only open when building) but tbf before I imagine it would crash since I never error handled any of this. You can also keep Log.Warn and add the message box in case the user auto clicks to close the MessageBox

shananas (approved):
I maintain there should be an actual message box popup since the user wont see Log.Warn without closing mod manager or doing a build command then opening the log text file but I wont have panacea settings be out of date longer

PR #1250, opened 2026-02-18
Fix Crash on Bootup if Save Directory does not exist.

TopazTK:
Adhoc fix, needs testing on EGS (if it even happens due to not having a bypass), but this will hold in the meantime.

// Deleted CodeRabbit summerization, we don't do that here.


Delta-473 (approved):
Approved

TopazTK:
Please revert the merge. I thought I have closed this PR yesterday but did not. I will open a new PR with a much more secure way of accomplishing this result.

PR #1262, opened 2026-06-22
Allow Choosing Branch for Mod in Mod Manager

gaithern:
This should allow choosing a branch apart from the default in the Mod Manager.  Useful for allowing others to test things without requiring the mod is updated or pointing the user to another repository.

### DIFF
--- a/OpenKh.Tools.ModsManager/Services/ModsService.cs
+++ b/OpenKh.Tools.ModsManager/Services/ModsService.cs
@@ -284,10 +284,10 @@ namespace OpenKh.Tools.ModsManager.Services
             string branchName = null)
         {
             var parts = repositoryName.Split('/');
-            if (parts.Length == 3)
+            if (parts.Length >= 3)
             {
                 if (string.IsNullOrWhiteSpace(branchName))
-                    branchName = parts[2];
+                    branchName = string.Join("/", parts.Skip(2));
                 repositoryName = $"{parts[0]}/{parts[1]}";
             }
 
@@ -633,25 +633,29 @@ namespace OpenKh.Tools.ModsManager.Services
             }
             else
             {
-                ClearOldMods(modName);
+                ClearOldModsCore(modName);
                 throw new ModMovedWithoutGameException(modName);
             }
         }));
 
-        public static Task ClearOldMods(string modName) => Task.Run(() => Handle(() =>
+        public static Task ClearOldMods(string modName) => Task.Run(() => Handle(() => ClearOldModsCore(modName)));
+
+        private static void ClearOldModsCore(string modName)
         {
             foreach (var game in new List<string> { "kh1", "kh2", "bbs", "Recom", "kh3d" })
             {
                 var modPath = GetModsGamePath(modName, game);
                 if (Directory.Exists(modPath))
-                    CleanModFiles(modPath);
+                    CleanModFilesCore(modPath);
             }
             var collectionPath = GetCollectionPath(modName);
             if (Directory.Exists(collectionPath))
-                CleanModFiles(collectionPath);
-        }));
+                CleanModFilesCore(collectionPath);
+        }
 
-        public static Task CleanModFiles(string modPath) => Task.Run(() => Handle(() =>
+        public static Task CleanModFiles(string modPath) => Task.Run(() => Handle(() => CleanModFilesCore(modPath)));
+
+        private static void CleanModFilesCore(string modPath)
         {
             foreach (var filePath in Directory.GetFiles(modPath, "*", SearchOption.AllDirectories))
             {
@@ -660,7 +664,7 @@ namespace OpenKh.Tools.ModsManager.Services
                     File.SetAttributes(filePath, attributes & ~FileAttributes.ReadOnly);
             }
             Directory.Delete(modPath, true);
-        }));
+        }
 
         public static Metadata GetMetadata(string modPath)
         {
--- a/OpenKh.Tools.ModsManager/Services/ModsService.cs
+++ b/OpenKh.Tools.ModsManager/Services/ModsService.cs
@@ -284,9 +284,10 @@ namespace OpenKh.Tools.ModsManager.Services
             string branchName = null)
         {
             var parts = repositoryName.Split('/');
-            if (parts.Length == 3 && string.IsNullOrWhiteSpace(branchName))
+            if (parts.Length == 3)
             {
-                branchName = parts[2];
+                if (string.IsNullOrWhiteSpace(branchName))
+                    branchName = parts[2];
                 repositoryName = $"{parts[0]}/{parts[1]}";
             }
 
@@ -328,17 +329,15 @@ namespace OpenKh.Tools.ModsManager.Services
                 switch (errorMessage)
                 {
                     case MessageBoxResult.Yes:
-                        Handle(() =>
-                        {
-                            MainViewModel.overwriteMod = true;
-                            ClearOldMods(repositoryName);
-                        });
+                        MainViewModel.overwriteMod = true;
+                        foreach (var existingModPath in modPaths.Where(Directory.Exists))
+                            await CleanModFiles(existingModPath);
+                        await ClearOldMods(repositoryName);
                         break;
                     case MessageBoxResult.No:
                         throw new ModAlreadyExistsExceptions(repositoryName);
                 }
             }
-            await Task.Delay(1); // fixes a bug where folder creation and deletions collide
             var modPath = modPaths[0];
             Directory.CreateDirectory(modPath);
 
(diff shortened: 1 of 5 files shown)


shananas:
Currently this PR breaks installing over a previously installed mod with this error.

image

not sure why this is happening but its consistent if i install a mod from github then immediately install it again after clicking yes to overwrite it i get this error even when stepping through in debug. the fix for this previously was the await Task.Delay(1) but its seems broken by this somehow.

also I believe the intention is to allow github inputs like `shananas/APCompanion/test-later` so `author/reponame/repobranch`

I did look into this previously but did not finish anything because my first step was to make a second input box on the UI for the user to type the branch into and if blank it would do main branch. personally I'd prefer if this PR did something similar with editing the UI since this is technically not how the branch URL looks and a separate input box would be preferred.

I wont deny it based on this but you will need to look into recreating the error I am seeing and fixing it if im the only one seeing it maybe its a me problem but if its happening to me i imagine itll happen to people when this is released causing headaches.

TopazTK:
> Currently this PR breaks installing over a previously installed mod with this error.
> 
> image
> not sure why this is happening but its consistent if i install a mod from github then immediately install it again after clicking yes to overwrite it i get this error even when stepping through in debug. the fix for this previously was the await Task.Delay(1) but its seems broken by this somehow.
> 
> also I believe the intention is to allow github inputs like `shananas/APCompanion/test-later` so `author/reponame/repobranch`
> 
> I did look into this previously but did not finish anything because my first step was to make a second input box on the UI for the user to type the branch into and if blank it would do main branch. personally I'd prefer if this PR did something similar with editing the UI since this is technically not how the branch URL looks and a separate input box would be preferred.
> 
> I wont deny it based on this but you will need to look into recreating the error I am seeing and fixing it if im the only one seeing it maybe its a me problem but if its happening to me i imagine itll happen to people when this is released causing headaches.

Couldn't he just do Directory.Exists(targetDir) => Directory.Delete(targetDir)?
Also this error is not specific to this PR, as the current OpenKH versions bug out if trying to install a mod that was previously present or incompletely deleted.

gaithern:
- Fixed the repositoryName/branch normalization bug (CodeRabbit): owner/repo/branch is now always normalized to owner/repo for the API/clone calls, even when a branch is also passed in explicitly.

- Fixed the overwrite-existing-mod crash: the old mod directory is now actually deleted (and awaited) before re-cloning on overwrite, instead of relying on a Task.Delay(1) that didn't address the real race. Verified by installing a mod, then reinstalling over it.

- Added a dedicated Branch textbox to the Install Mod dialog, wired to the existing (previously unused) BranchName property. The owner/repo/branch shorthand still works too; if both are provided, the textbox takes priority.

shananas:
> > Currently this PR breaks installing over a previously installed mod with this error.
> > image
> > not sure why this is happening but its consistent if i install a mod from github then immediately install it again after clicking yes to overwrite it i get this error even when stepping through in debug. the fix for this previously was the await Task.Delay(1) but its seems broken by this somehow.
> > also I believe the intention is to allow github inputs like `shananas/APCompanion/test-later` so `author/reponame/repobranch`
> > I did look into this previously but did not finish anything because my first step was to make a second input box on the UI for the user to type the branch into and if blank it would do main branch. personally I'd prefer if this PR did something similar with editing the UI since this is technically not how the branch URL looks and a separate input box would be preferred.
> > I wont deny it based on this but you will need to look into recreating the error I am seeing and fixing it if im the only one seeing it maybe its a me problem but if its happening to me i imagine itll happen to people when this is released causing headaches.
> 
> Couldn't he just do Directory.Exists(targetDir) => Directory.Delete(targetDir)? Also this error is not specific to this PR, as the current OpenKH versions bug out if trying to install a mod that was previously present or incompletely deleted.

I did not make the Delay(1) addition that was needed for the mod collections PR.

When I gave users the ability to overwrite the mod in my PR with the dialogue box etc. I never ran into any error of the folder existing once I had fixed it. I do not remember what i did id have to look at my PR and the code paths very well could've changed since then especially with the collections PR. I also saw no reports from when I added the option to overwrite (again once i fixed all the errors which happened before the PR was made).
I am not saying they didn't happen people don't report things all the time but I saw no reports.
I also saw no reports with the collections PR of this folder issue once they had a fix. I forget if I commented on that PR with this error first but I do recall asking why are we delaying 1 ms and the response was to prevent this folder issue so I said ok.

basically before this PR any folder conflict like this didn't show in my testing and I saw no reports.

Also since its a github clone we need to change some folder/file permissions before deleting them otherwise an error happens which is what the called method does change permissions then deletes the files.

TopazTK:
> > > Currently this PR breaks installing over a previously installed mod with this error.
> > > image
> > > not sure why this is happening but its consistent if i install a mod from github then immediately install it again after clicking yes to overwrite it i get this error even when stepping through in debug. the fix for this previously was the await Task.Delay(1) but its seems broken by this somehow.
> > > also I believe the intention is to allow github inputs like `shananas/APCompanion/test-later` so `author/reponame/repobranch`
> > > I did look into this previously but did not finish anything because my first step was to make a second input box on the UI for the user to type the branch into and if blank it would do main branch. personally I'd prefer if this PR did something similar with editing the UI since this is technically not how the branch URL looks and a separate input box would be preferred.
> > > I wont deny it based on this but you will need to look into recreating the error I am seeing and fixing it if im the only one seeing it maybe its a me problem but if its happening to me i imagine itll happen to people when this is released causing headaches.
> > 
> > 
> > Couldn't he just do Directory.Exists(targetDir) => Directory.Delete(targetDir)? Also this error is not specific to this PR, as the current OpenKH versions bug out if trying to install a mod that was previously present or incompletely deleted.
> 
> I did not make the Delay(1) addition that was needed for the mod collections PR.
> 
> When I gave users the ability to overwrite the mod in my PR with the dialogue box etc. I never ran into any error of the folder existing once I had fixed it. I do not remember what i did id have to look at my PR and the code paths very well could've changed since then especially with the collections PR. I also saw no reports from when I added the option to overwrite (again once i fixed all the errors which happened before the PR was made). I am not saying they didn't happen people don't report things all the time but I saw no reports. I also saw no reports with the collections PR of this folder issue once they had a fix. I forget if I commented on that PR with this error first but I do recall asking why are we delaying 1 ms and the response was to prevent this folder issue so I said ok.
> 
> basically before this PR any folder conflict like this didn't show in my testing and I saw no reports.
> 
> Also since its a github clone we need to change some folder/file permissions before deleting them otherwise an error happens which is what the called method does change permissions then deletes the files.

I didn't say you made it or no, I was just asking if this fix was a possibility. GICU has insured it was.
I also did not say the bug was collection exclusive.

I did not say anything to insult or offend, I swear.

shananas:
I take no offense just wanted to clarify the history of replacing a mod that's already installed since I made it and when it was implemented it was working since its been through a couple changes.
Main point being you cant just delete the directory due to certain file locks placed on files when cloning from github which needs extra code to unlock which does exist in the called functions.

TopazTK:
> I take no offense just wanted to clarify the history of replacing a mod that's already installed since I made it and when it was implemented it was working since its been through a couple changes. Main point being you cant just delete the directory due to certain file locks placed on files when cloning from github which needs extra code to unlock which does exist in the called functions.

I have never encountered this, aren't we using Octokit?

OpenKh.Tools.ModsManager/Services/ModsService.cs
```diff
@@ -279,24 +280,43 @@ public static void InstallModFromZip(
...
+            }
 
-                progressOutput?.Invoke($"Fetching file {ModMetadata} from {branchName}");
-                isValidMod = await RepositoryService.IsFileExists(repositoryName, branchName, ModMetadata);
+            if (!string.IsNullOrWhiteSpace(branchName))
+            {
+                progressOutput?.Invoke($"Fetching file {ModMetadata} from branch {branchName}");
+                var isValidMod = await RepositoryService.IsFileExists(repositoryName, branchName, ModMetadata);
+                if (!isValidMod)
+                    throw new ModNotValidException(repositoryName);
```
shananas: To separate a possible bad branch from a bad repo entirely this should be a new error i did a quick mock up mostly copying the error you are currently using. changed this to `throw new BranchNotValidException(repositoryName, branchName);`

Then in Exceptions.cs (since i cant review a file you didn't modify I added this as a new exception. You can adjust the error message if you want I think its fine as is but I don't care either way

```
    public class BranchNotValidException : Exception
    {
        public string ModName { get; }

        public string BranchName { get; }

        public BranchNotValidException(string modName, string branchName) :
            base($"The branch '{branchName}' is not valid for '{modName}'.\nOr The mod '{modName}' does not contain a valid OpenKH compatible mod due to the missing 'mod.yml' file.")
        {
            ModName = modName;
            BranchName = branchName;
        }
    }
```

shananas (changes requested):
Just my one requested change and everything works for me and is good

In my head ideally branch name would be a semi small box on the right side of where you put in the repo but ideally I think it would be hard to make look good so that's fine as is.

shananas (approved):

PR #1263, opened 2026-06-23
Fix OpenKh.Command.Bbsa extraction logic

ktheticdev:
This fixes extraction by including files that were not extracted previously, thus fixing repacking, too. Specifically this targets the HE_205.exa file in KH BBS FM PSP release.


shananas (approved):
Based on Thamstras comment this won't break anything LGTM

PR #1269, opened 2026-08-01
[ModsManager] Make it possible to fetch mods from any Git platform of choice.

TopazTK:
Exactly what it says on the tin. This makes it possible for the ModsManager to fetch and download mods from other Git vendors, breaking its sole reliability on GitHub.

## Why?
Why not! I won't be using GitHub much longer due to... issues, and I believe more and more people will choose not to also. So I believe this is a legit requirement in the ModsManager.

## Differences in Performance
Fetching the "mod.yml" file to check if the mod is valid takes longer on non-GitHub platforms due to having to use LibGit2Sharp's methods to fetch the file instead of relying on REST API calls. This is necessary because literally every Git platform does it's job differently. Other than that, non-GitHub platforms seem to fetch the mods **faster** (KH-ReFined/KH2-MAIN cloned from Codeberg in 10 seconds (exc. mod fetch). Same mod cloned from GitHub in 7 minutes and 35 seconds.)

## Screenshots
image
image
image
image

## Testing
All the necessary testing was done on my end and I cannot see any issues with non-GitHub platforms nor any regression with GitHub. However, y'all are free to test and report issues to me as time goes.

### DIFF
--- a/OpenKh.Tools.ModsManager/Services/ModsService.cs
+++ b/OpenKh.Tools.ModsManager/Services/ModsService.cs
@@ -121,11 +121,12 @@ namespace OpenKh.Tools.ModsManager.Services
             bool isLuaFile,
             Action<string> progressOutput = null,
             Action<float> progressNumber = null,
+            string platformUrl = null,
             string branchName = null)
         {
             if (!isZipFile && !isLuaFile)
             {
-                return Task.Run(() => InstallModFromGithub(name, progressOutput, progressNumber, branchName));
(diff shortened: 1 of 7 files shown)


solt-frfr:
This would be a great feature to have! More options are always better and it's still important to be able to install projects that have decided to stop using GitHub as their place to host it.

From what I can tell, the code looks just fine and seems to work great

TopazTK:
Made changes @shananas requested on the OpenKH server.

shananas (approved):

PR #1270, opened 2026-08-01
[ModsManager] Execute a Git install if ZIP has repository information.

TopazTK:
Self-Explanatory, this is only a requirement for stuff like Nexus Mods and if you also plan on updating your mods.

New field in YAML called "repository". This field must have a valid Git Repository link. [Yes, even with GitHub, a full link is necessary.]

image


TopazTK:
Closed until we can find a better solution for this.

PR #1274, opened 2026-08-13
Fix ModCollectionPath folder nesting

tryashtar:
By default, the tool looks for mods in `<root>/mods`.

You can change the `modCollectionPath` option in `mods-manager.yml` to make it look somewhere else instead. However, currently this makes it look in `<path>/mods`, instead of just `<path>`.

This PR fixes this without affecting the default `<root>/mods` location.

### DIFF
--- a/OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
+++ b/OpenKh.Tools.ModsManager/Services/ConfigurationService.cs
@@ -273,7 +273,7 @@ namespace OpenKh.Tools.ModsManager.Services
 
         public static string GameModPath
         {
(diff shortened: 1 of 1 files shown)


shananas:
Both the installed mods and built mod folder use modCollectionsPath in get so this happens to separate installed mods into one subfolder and built mods into another im going to fix this bug and while im at it rename these variables to be less confusing. Ill add this change to that for both mods and mod.

image

as seen here the ahh folder isnt used at all. well i deleted it but it stayed empty build mods went into the mod folder

tryashtar:
Hm, yeah, the `GameModPath` has a typo in the getter. I'll push a fix for that too.

shananas (approved):
This technically introduces a bug if ModCollectionsPath and GameModsPath are the same but im fixing it while renaming these variables to be less confusing

PR #1276, opened 2026-08-15
Rename some configurationservicevariables.

shananas:
Also ensures if both the path for the mods and mod are the same sub-folders are still created to prevent building mods causing overwrites and errors.


shananas:
@coderabbitai review

shananas:
@coderabbitai review

PR #1277, opened 2026-08-16
Adds kh1ardresource method

gaithern:
Small one.  This adds the kh1ardresouce method.

KH1 `*.ard` files include a list of model/animation resource filenames that each room loads/uses.
These references are in plain ASCII and are pointed to via offset at the beginning of the binary.

We can patch these by reading that offset and replacing the plaintext references by index since they all have the same size capacity.

I want to use this for KH1 enemy randomizer without replacing the entire `*.ard` binaries, not sure what else this would be used for.

### DIFF
--- a/docs/tool/GUI.ModsManager/creatingMods.md
+++ b/docs/tool/GUI.ModsManager/creatingMods.md
@@ -423,22 +423,33 @@ This method rewrites individual slots in place, so the rest of the binary is lef
 Use it to swap a model or animation set without having to `copy` the whole multi-megabyte
 `.ard` into your mod, which would clobber any other mod's changes to that file.
 
-Unlike most methods, this one takes no `source`. The replacements are declared inline with
-`replacements`, keyed by slot index:
+Asset Example
 
 ```
 - name: tw01.ard
   method: kh1ardresource
-  replacements:
-    0: xa_al_9999.mdls
-    1: xa_al_9999.mset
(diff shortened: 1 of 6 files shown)


docs/tool/GUI.ModsManager/creatingMods.md
```diff
@@ -413,6 +414,32 @@ Asset Example
...
+
+```
+- name: tw01.ard
+  method: kh1ardresource
+  replacements:
+    0: xa_al_9999.mdls
+    1: xa_al_9999.mset
+    8: tw_6100.moa
+    9: tw_6100.moa.mset
+```
```
shananas, line 433: Ideally for consistency this new method would still pull from another file like `tw01.yml` (to match the example here) and youd write the mod.yml something like

```
- name: tw01.ard
  method: kh1ardrescource
  source
    - name: tw01.yml (or filepath/tw01.yml)
```

This would keep the behavior that the mod.yml is only really edited to add/remove an edit entirely (mostly) and any updates to edits happen within the file linked to by the mod.yml

gaithern: Oh okay gotcha - I'll make that change

gaithern: Should be resolved now

shananas (approved):

PR #1278, opened 2026-08-19
Fix all Git Problems on ModsManager

TopazTK:
This PR does the following:
- Add feedback if fetching the "mod.yml" from another Git service.
- Make it so that you can use an @ modifier for platform name instead of declaring it on the other box [Ex. KH-ReFined/KH2-MAIN@codeberg.org]
- Speed up fetching said repository from another Git service. (15 seconds Codeberg fetching "KH-ReFined/KH2-MAIN")
- Ensure "gitfetch" is permission fixed and deleted on initialization and mod installations.
- Ensure that all permissions for every single mod installed is updated on initialization. This should, hopefully, fix many issues regarding mods.


shananas (approved):
works on my machine ¯\_(ツ)_/¯

PR #1282, opened 2026-08-21
Dont show patching menu unless user specifically enables it.

shananas:
At this point I don't think many people are using OpenKH for patching but still want to keep mod managers abilities to do so.

The main thing this solves by not showing patching menu if panacea isn't installed is its very common for KH2 randomizers people get the patching menu to show either by not installing panacea or some other way and patch their game files. This causes extra troubleshooting and time in verifying game files/using the Restore button.

Not sure how far away Yokimitsuros mod manager port to avalonia is. This is a change that will save many more basic users just trying to play rando than work for people who still use patching to check the menu option. Ive considered moving Enable Patching to setting menu but I fear basic users will for some reason check that if its under settings.


shananas:
@coderabbitai review

kenjiuno (approved):

